PR #319, opened 2022-04-11
Add a test pool with 4 decimals (+ some testing-related refactoring)

lbeder:


test/deployment/000025-create-test-network.ts
```diff
@@ -30,15 +30,17 @@ describeDeployment(
             [InstanceName.TestToken2]: toWei(500_000),
             [InstanceName.TestToken3]: toWei(1_000_000),
             [InstanceName.TestToken4]: toWei(2_000_000),
-            [InstanceName.TestToken5]: toWei(3_000_000)
+            [InstanceName.TestToken5]: toWei(3_000_000),
+            [InstanceName.TestToken6]: toWei(100_000)
```
yudilevi, line 34: with 4 decimals that would be a huge deposit - should we use a smaller number here?

lbeder: Hmmm... yeah, let's reduce it a little bit

yudilevi (approved):

PR #321, opened 2022-04-12
Simplify trading fee amount handling by making sure that it's always the full amount across all components of the system

lbeder:
…the full amount across all components of the system


contracts/network/BancorNetwork.sol
```diff
@@ -1247,7 +1247,8 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
                 params.limit
             );
 
-        // if the target token is BNT, notify the BNT pool on collected fees
+        // if the target token is BNT, notify the BNT pool on collected fees (which shouldn't include the network fee
+        // amount, so we have to deduct it explicitly from the full trading fee amount)
```
barakman, line 1251: Unrelated, but we can take the opportunity and call this function only if the amount is larger than zero, in order to save the gas cost of an external delegate call (in addition, I'd remove the 'amount == 0' check inside `onFeesCollected`, because it is used there only for gas optimization).
That being said, the whole thing is really worth the effort, if we actually expect this amount to ever be zero.

barakman (approved):
I'm assuming that by the least, the Financial Verification test in BancorNetwork.ts should fail at this point; perhaps best to fix it as part of this PR.

yudilevi (dismissed):

contracts/network/BancorNetwork.sol
lbeder: The trading fee is extremely unlikely to be 0. It can only happen if it's set to 0 or the network fee amount is set to 100%.

lbeder:
> > > I'm assuming that by the least, the Financial Verification test in BancorNetwork.ts should fail at this point; perhaps best to fix it as part of this PR.
> > 
> > 
> > All financial verification tests are passing, so it means that there's a bug in them. Can you please take a look?
> 
> These tests merely run a set of user operations (deposit/withdraw/trade), and then verify that the state of the system is as expected. If they all complete successfully, then it means that this change hasn't made any impact on the specific flow in each one of these tests (at least with respect to the state of the system being measured after every user operation).

Yes, one of the nightly scenarios is affected

yudilevi (approved):

contracts/network/BancorNetwork.sol
barakman: Or the trading amount is really small (such that the trading fee fraction of it becomes zero)...
Resolving.

PR #329, opened 2022-04-14
Ensure to remove an inactive program, with no remaining rewards, from providers' program list only if their stakes are zero as well

lbeder:
… providers' program list only if their stakes are zero as well


yudilevi (approved):

barakman (approved):
Minor:
We've been using 'rewards' (plural) conclusively up until now.
This change introduces 'reward' (singular) into our terminology.

PR #344, opened 2022-04-17
Create a stateless logical model for all flows

barakman:
Create "single point of truth" logic scripts (model) that can be utilized by different systems/3rd parties.

This is an initial draft.


yudilevi:
Will revisit in the future

PR #345, opened 2022-04-17
Maintain the remaining rewards amount and use it as a sanity check when claiming rewards

lbeder:


contracts/rewards/StandardRewards.sol
```diff
@@ -48,7 +48,6 @@ contract StandardRewards is IStandardRewards, ReentrancyGuardUpgradeable, Utils,
     }
 
     struct RewardData {
-        Token pool;
```
yudilevi, line 51: just a remnant from a previous iteration?

contracts/rewards/StandardRewards.sol
```diff
@@ -739,7 +735,21 @@ contract StandardRewards is IStandardRewards, ReentrancyGuardUpgradeable, Utils,
             }
 
             ClaimData memory claimData = _claimRewards(provider, p);
-            rewardData.amount += claimData.reward;
+
+            if (claimData.reward > 0) {
+                uint256 remainingRewards = p.remainingRewards;
+
+                // a sanity check that the reward amount doesn't exceed the remaining rewards per program
+                if (remainingRewards < claimData.reward) {
```
yudilevi, line 743: this isn't needed, the following subtraction ensures that it'll revert - is that here just for the error message?
if so, I wouldn't bother, since it's an edge case anyway, not an error that anyone should ever see unless there's a bug :)

lbeder: Yeah, I wanted to have an explicit revert reason if this ever happens (not that it should)

yudilevi: 👍

lbeder: I'm updating the revert reason, though.

yudilevi (approved):

contracts/rewards/StandardRewards.sol
```diff
@@ -370,7 +370,8 @@ contract StandardRewards is IStandardRewards, ReentrancyGuardUpgradeable, Utils,
             isEnabled: true,
             startTime: startTime,
             endTime: endTime,
-            rewardRate: totalRewards / (endTime - startTime)
+            rewardRate: rewardRate,
+            remainingRewards: rewardRate * (endTime - startTime)
```
barakman, line 374: Is your purpose here to get rid of the remainder in `totalRewards / (endTime - startTime)`?

If not, then you may as well keep the code as before and just add:
```
remainingRewards: totalRewards
```

barakman (changes requested):

contracts/rewards/StandardRewards.sol
lbeder: Yes, the rationale is to make sure that `remainingRewards` is equal to the rewards which are actually going to be distributed (`rewardRate * (endTime - startTime)`), which is usually a wei awei (🥁🥁) from `totalRewards`

barakman (approved):

PR #346, opened 2022-04-17
Remove (solvable) cyclic dependencies for etherscan/tenderly verification 

lbeder:


yudilevi (approved):

contracts/utility/Fraction.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: SEE LICENSE IN LICENSE
```
barakman, line 1: Perhaps consider reinstating the (globally-declared) `Types` module, to include these two types along with the ones moved to `MathEx` (`Sint256` and `Uint512`).

barakman (approved):

contracts/utility/Fraction.sol
lbeder: Generally, we should try to keep types closer to their implementation, it's just in this case, we had both `FractionLibrary.sol` and `MathEx.sol` referencing each other, and this was a quick fix for it

PR #353, opened 2022-04-18
Add coverage report (+ slightly improved the coverage)

lbeder:


test/network/NetworkSettings.ts
```diff
@@ -179,13 +181,21 @@ describe('NetworkSettings', () => {
                 const reserveToken2 = await createTestToken();
                 expect(await networkSettings.isTokenWhitelisted(reserveToken.address)).to.be.false;
                 expect(await networkSettings.isTokenWhitelisted(reserveToken2.address)).to.be.false;
+                expect(await networkSettings.protectedTokenWhitelist()).not.to.have.members([
```
barakman, line 184: `to.be.empty` would be a little more restrictive.

test/network/NetworkSettings.ts
```diff
@@ -179,13 +181,21 @@ describe('NetworkSettings', () => {
...
+                    reserveToken2.address
+                ]);
 
                 const res = await networkSettings.addTokensToWhitelist([reserveToken.address, reserveToken2.address]);
                 await expect(res).to.emit(networkSettings, 'TokenAddedToWhitelist').withArgs(reserveToken.address);
                 await expect(res).to.emit(networkSettings, 'TokenAddedToWhitelist').withArgs(reserveToken2.address);
 
                 expect(await networkSettings.isTokenWhitelisted(reserveToken.address)).to.be.true;
                 expect(await networkSettings.isTokenWhitelisted(reserveToken2.address)).to.be.true;
+                expect(await networkSettings.protectedTokenWhitelist()).to.have.members([
```
barakman, line 195: `to.deep.equal` would be a little more restrictive.

test/network/NetworkSettings.ts
```diff
@@ -211,11 +221,13 @@ describe('NetworkSettings', () => {
 
             it('should remove a token', async () => {
                 expect(await networkSettings.isTokenWhitelisted(reserveToken.address)).to.be.true;
+                expect(await networkSettings.protectedTokenWhitelist()).to.include(reserveToken.address);
 
                 const res = await networkSettings.removeTokenFromWhitelist(reserveToken.address);
                 await expect(res).to.emit(networkSettings, 'TokenRemovedFromWhitelist').withArgs(reserveToken.address);
 
                 expect(await networkSettings.isTokenWhitelisted(reserveToken.address)).to.be.false;
+                expect(await networkSettings.protectedTokenWhitelist()).not.to.include(reserveToken.address);
```
barakman, line 230: Semantic: I'd use `to.not.include` (3 places in this file, one of which is not part of this PR).

barakman (approved):

test/network/NetworkSettings.ts
lbeder: The test only checks the difference between adding/removing specific tokens, so it's unnecessary also to assume that the list is empty (which is incidentally the case)

test/network/NetworkSettings.ts
lbeder: The test only checks the difference between adding/removing specific tokens, so it's unnecessary also to assume that the list was empty (which is incidentally the case)

test/network/NetworkSettings.ts
lbeder: Sure

lbeder: Actually, it works either way, so let's not waste time on this

PR #361, opened 2022-04-21
Process rewards automatic trigger

barakman:


contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -305,10 +320,51 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
...
+        });
+    }
+
+    /**
+     * @inheritdoc IAutoCompoundingRewards
+     */
+    function trigger() external nonReentrant {
+        address[] memory values = _pools.values();
+        uint256 numOfPools = values.length;
+        uint256 startIndex = _nextTriggerIndex;
```
yudilevi, line 346: no need for startIndex, can use _nextTriggerIndex directly

contracts/rewards/interfaces/IAutoCompoundingRewards.sol
```diff
@@ -78,6 +83,20 @@ interface IAutoCompoundingRewards is IUpgradeable {
...
+     * @dev sets the number of programs to process the rewards of, when auto-trigerred
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
+     */
+    function setAutoTriggerCount(uint256 autoTriggerCount) external;
+
+    /**
+     * @dev processes program rewards for several pools
```
yudilevi, line 96: maybe `processes program rewards based on internal logic, without requiring any input`?

yudilevi (commented):
looks fine to me, though I'd also add a check inside `_processRewards` to skip the current pool if it was processed in the past 1h (according to the spec)

contracts/rewards/AutoCompoundingRewards.sol
barakman: It's being used inside the loop, so better cache it first.

contracts/rewards/interfaces/IAutoCompoundingRewards.sol
barakman: I don't see any added value in 'without requiring any input', since it is already obvious from the function prototype.

yudilevi: it's the concept and the rationale for its existence

contracts/rewards/AutoCompoundingRewards.sol
yudilevi: oh makes sense

contracts/rewards/interfaces/IAutoCompoundingRewards.sol
barakman: Done

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -305,10 +322,51 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
...
      * @inheritdoc IAutoCompoundingRewards
      */
     function processRewards(Token pool) external nonReentrant {
+        _processRewards(pool, false);
+    }
+
+    /**
+     * @dev processes the rewards of a given pool
+     */
+    function _processRewards(Token pool, bool skip) private {
```
yudilevi, line 369: skip implies that if true, the request is going to be skipped

barakman: fixed

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -366,7 +366,7 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
     /**
      * @dev processes the rewards of a given pool
      */
-    function _processRewards(Token pool, bool skip) private {
+    function _processRewards(Token pool, bool skipRecentlyProcessedPrograms) private {
```
yudilevi, line 369: maybe `forceProcessing` and reverse the conditions?

lbeder: I agree with @barakman. It's something that should better require an opt-in (e.g., explicit `true`)

yudilevi: sure, so maybe just a shorter name? `skipRecent`?

barakman: done

yudilevi (dismissed):

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -42,6 +42,8 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
 
     error InsufficientFunds();
 
+    uint16 private constant AUTO_TRIGGER_MIN_TIME_DELTA = 1 hours;
```
lbeder, line 45: I'd add a comment here

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -89,6 +94,11 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
      */
     event ProgramEnabled(Token indexed pool, bool status, uint256 remainingRewards);
 
+    /**
+     * @dev triggered when the auto trigger count is updated
+     */
+    event AutoTriggerCountUpdated(uint256 prevAutoTriggerCount, uint256 newAutoTriggerCount);
```
lbeder, line 100: @yudilevi is this necessary? It costs more gas

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -281,10 +298,51 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
...
+            prevAutoTriggerCount: prevAutoTriggerCount,
+            newAutoTriggerCount: newAutoTriggerCount
+        });
+    }
+
+    /**
+     * @inheritdoc IAutoCompoundingRewards
+     */
+    function trigger() external nonReentrant {
+        address[] memory values = _pools.values();
```
lbeder, line 322: Keep in mind that this loads all the `_pools` to the memory and if there are too many (e.g., tens of thousands) - this will exceed the block gas limit. This isn't an issue, but we should still consider not loading all the pools to reduce the gas cost of this function, especially when `values` are only accessed by index anyway...

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -281,10 +298,51 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
...
+        uint256 numOfPools = values.length;
+        uint256 startIndex = _nextTriggerIndex;
+        uint256 count = Math.min(numOfPools, _autoTriggerCount);
+
+        for (uint256 i = 0; i < count; i++) {
+            uint256 index = (startIndex + i) % numOfPools;
+            _processRewards(Token(values[index]), true);
```
lbeder, line 329: Perhaps `_processRewards` should return whether the program was processed or not (== skipped) and then it shouldn't be taken into account and the next program should be considered instead?

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -281,10 +298,51 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
         emit ProgramEnabled({ pool: pool, status: status, remainingRewards: p.remainingRewards });
     }
 
+    /**
+     * @inheritdoc IAutoCompoundingRewards
+     */
+    function setAutoTriggerCount(uint256 newAutoTriggerCount) external onlyAdmin {
+        uint256 prevAutoTriggerCount = _autoTriggerCount;
```
lbeder, line 305: `newAutoTriggerCount` should be at least greater than 0, no?

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -89,6 +94,11 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
      */
     event ProgramEnabled(Token indexed pool, bool status, uint256 remainingRewards);
 
+    /**
+     * @dev triggered when the auto trigger count is updated
```
lbeder, line 98: Nitpick: I'd change the comment to `auto-trigger`, like you did in other places

contracts/rewards/interfaces/IAutoCompoundingRewards.sol
```diff
@@ -93,6 +98,20 @@ interface IAutoCompoundingRewards is IUpgradeable {
...
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
+     */
+    function setAutoTriggerCount(uint256 autoTriggerCount) external;
+
+    /**
+     * @dev processes program rewards based on internal logic, without requiring any input
+     */
+    function trigger() external;
```
lbeder, line 113: I'm not in love with the generic `trigger` name. Maybe something like `autoProcessRewards` or  `triggerProcessRewards`?

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -686,6 +711,83 @@ describe('AutoCompoundingRewards', () => {
...
+                            for (let i = 0; i < 5; i++) {
+                                await autoCompoundingRewards.setTime(
+                                    Math.floor(START_TIME + (programDuration[distributionType] * 2 ** i) / (2 ** i + 1))
+                                );
+                                const res = await autoCompoundingRewards.trigger();
+                                await expect(res).to.emit(autoCompoundingRewards, 'RewardsDistributed');
+                            }
+                        });
+
+                        it('should not distribute tokens again after less than 1 hour', async () => {
```
lbeder, line 772: `1 hour` may change, so I'd make this description more general and also move `60 * 60` (`hours(1)`) to a `AUTO_TRIGGER_MIN_TIME_DELTA` constant (Constants.ts) instead

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -686,6 +711,83 @@ describe('AutoCompoundingRewards', () => {
...
+                            for (let i = 0; i < Math.ceil(setups.length / AUTO_TRIGGER_COUNT); i++) {
+                                const res = await autoCompoundingRewards.trigger();
+                                await expect(res).to.emit(autoCompoundingRewards, 'RewardsDistributed');
+                            }
+                            await autoCompoundingRewards.setTime(
+                                Math.floor(START_TIME + programDuration[distributionType] / 2) + 60 * 60 - 1
+                            );
+                            for (let i = 0; i < Math.ceil(setups.length / AUTO_TRIGGER_COUNT) + 1; i++) {
+                                const res = await autoCompoundingRewards.trigger();
+                                await expect(res).not.to.emit(autoCompoundingRewards, 'RewardsDistributed');
```
lbeder, line 785: Can you add a test that only some programs are skipped and not only them? You can add a `TestAutoCompoundingRewards`-only function that allows you to override `prevDistributionTimestamp` (or to manually create a few "holes" by manually triggering their rewards processing)

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -66,8 +68,11 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
     // a set of all pools that have a rewards program associated with them
     EnumerableSetUpgradeable.AddressSet private _pools;
 
+    uint256 private _autoTriggerCount;
```
lbeder, line 71: I'd add a few comments here as well

lbeder (changes requested):

contracts/rewards/AutoCompoundingRewards.sol
yudilevi: the event? it's an admin function, we always add events when updating admin settings

contracts/rewards/AutoCompoundingRewards.sol
yudilevi: good idea + will make the cost/reward more consistent

contracts/rewards/interfaces/IAutoCompoundingRewards.sol
yudilevi: `autoProcess`?

contracts/rewards/AutoCompoundingRewards.sol
lbeder: Oh, yeah. Resolving.

contracts/rewards/interfaces/IAutoCompoundingRewards.sol
lbeder: `autoProcess` works too

contracts/rewards/AutoCompoundingRewards.sol
barakman: Well, the alternative - calling `_pools.at` - yields an **internal** function call, because `EnumerableSetUpgradeable.AddressSet` is used as an internal library in this contract.

On the other hand, it essentially passes `_pools` to function `at`, which might yield an additional storage-read per iteration (I'm only 50% sure of that TBH).

So possibly, changing to this scheme might also increase the gas cost.
In fact, there's a good chance that the internal function call per iteration alone (i.e., even without the additional storage-read per iteration) will increase the gas cost in the same manner as reading all the values into memory (at least in the specific case when `_autoTriggerCount` is as large as the number of pools).

We probably need to do some measurements, but I overall tend to accept this suggestion so fixed; please re-review.

contracts/rewards/AutoCompoundingRewards.sol
barakman: IMO, no. It's a DAO function, so it may as well give full flexibility (in this case, a kind of 'auto-trigger disablement').

contracts/rewards/AutoCompoundingRewards.sol
barakman: done

contracts/rewards/interfaces/IAutoCompoundingRewards.sol
barakman: done (pretty massive rename)

test/rewards/AutoCompoundingRewards.ts
barakman: done

contracts/rewards/AutoCompoundingRewards.sol
barakman: done

contracts/rewards/AutoCompoundingRewards.sol
barakman: done

test/rewards/AutoCompoundingRewards.ts
barakman: It's already there (one test above this one).

contracts/rewards/AutoCompoundingRewards.sol
barakman: done

contracts/rewards/interfaces/IAutoCompoundingRewards.sol
```diff
@@ -93,6 +98,20 @@ interface IAutoCompoundingRewards is IUpgradeable {
      */
     function enableProgram(Token pool, bool status) external;
 
+    /**
+     * @dev sets the number of programs to process the rewards for, when auto-trigerred
```
yudilevi, line 102: should probably say "auto processed"

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -634,6 +639,31 @@ describe('AutoCompoundingRewards', () => {
                     });
                 });
 
+                describe('setting auto-process count', () => {
```
yudilevi, line 642: missing test for 0 input

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -686,6 +716,85 @@ describe('AutoCompoundingRewards', () => {
                     }
                 });
 
+                describe('auto-process rewards', () => {
+                    const AUTO_PROCESS_COUNT = 3;
+
+                    const setups = [
```
yudilevi, line 722: can have an array of token symbols and initialize the rest in the loop

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -686,6 +716,85 @@ describe('AutoCompoundingRewards', () => {
...
+
+                        await autoCompoundingRewards.setAutoProcessCount(AUTO_PROCESS_COUNT);
+                    });
+
+                    if (distributionType === RewardsDistributionType.Flat) {
+                        it('should distribute all tokens', async () => {
+                            await autoCompoundingRewards.setTime(programEndTime[distributionType]);
+                            for (let i = 0; i < Math.ceil(setups.length / AUTO_PROCESS_COUNT); i++) {
+                                const res = await autoCompoundingRewards.autoProcess();
+                                await expect(res).to.emit(autoCompoundingRewards, 'RewardsDistributed');
```
yudilevi, line 761: shouldn't we check that the event was emitted the correct number of times in this case (after each autoProcess call)?

yudilevi (changes requested):

contracts/rewards/AutoCompoundingRewards.sol
barakman: But nevertheless... done

test/rewards/AutoCompoundingRewards.ts
barakman: Yes, but I wasn't really sure how to count them; probably something like `expect(res).to.emit(...).times(10)`, but I'm not sure if there's a syntax for that; I'll check it out...

contracts/rewards/interfaces/IAutoCompoundingRewards.sol
barakman: There's a spelling error in `trigerred`, which is why it hasn't been replaced with 'auto-processed' earlier on;
We need to consider the general naming convention in this context anyway (`processRewards` vs `autoProcess`).

test/rewards/AutoCompoundingRewards.ts
barakman: added

contracts/rewards/interfaces/IAutoCompoundingRewards.sol
barakman: fixed

test/rewards/AutoCompoundingRewards.ts
barakman: These are just random amounts that I've picked.
Initializing them in a loop means that we will not be able to easily change or add more of them.

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -639,32 +639,50 @@ describe('AutoCompoundingRewards', () => {
...
-                describe('setting auto-process count', () => {
-                    it('should revert when a non-admin attempts to set the auto-process count', async () => {
-                        await expect(autoCompoundingRewards.connect(user).setAutoProcessCount(1)).to.be.revertedWith(
-                            'AccessDenied'
+                describe('the number of programs to auto-process the rewards for', () => {
+                    it('should initially be larger than zero', async () => {
+                        expect(await autoCompoundingRewards.autoProcessRewardsCount()).to.be.gt(0);
+                    });
+
+                    it('cannot be set by the admin to zero', async () => {
```
yudilevi, line 647: should be something like `should revert if set to a 0 value`, otherwise it implies that a non admin can set it to 0

test/rewards/AutoCompoundingRewards.ts
yudilevi: ok, though look at the numbers now

barakman: Yeah, I wasn't picking "that randomly". But the idea is that they are independent of each other.
Computing them in a loop would some dependency which is not required (nor desired) in this case.

test/rewards/AutoCompoundingRewards.ts
barakman: fixed

test/rewards/AutoCompoundingRewards.ts
yudilevi: sounds good

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -281,26 +304,80 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
...
+        }
+
+        if (skipRecent && currTime < p.prevDistributionTimestamp + AUTO_PROCESS_REWARDS_MIN_TIME_DELTA) {
+            return false;
         }
 
         uint256 tokenAmountToDistribute = _tokenAmountToDistribute(p, currTime);
         if (tokenAmountToDistribute == 0) {
-            return;
+            return false;
```
lbeder, line 375: IMHO it should return `true` here, since it does indeed like a program was processed (just didn't yield any rewards; i.e., it wasn't actually skipped)

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -281,26 +304,80 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
...
+
     /**
      * @inheritdoc IAutoCompoundingRewards
      */
     function processRewards(Token pool) external nonReentrant {
+        _processRewards(pool, false);
+    }
+
+    /**
+     * @dev processes the rewards of a given pool
```
lbeder, line 358: "and returns whether the rewards processing was successfully triggered or skipped" (or something like that)

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -634,7 +639,50 @@ describe('AutoCompoundingRewards', () => {
                     });
                 });
 
-                describe('processing rewards', () => {
+                describe('the number of programs to auto-process the rewards for', () => {
```
lbeder, line 642: `describe` should describe the test group and not the test case itself

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -634,7 +639,50 @@ describe('AutoCompoundingRewards', () => {
...
+                    it('should initially be larger than zero', async () => {
+                        expect(await autoCompoundingRewards.autoProcessRewardsCount()).to.be.gt(0);
```
lbeder, line 644: This should be part of the `construction` test group

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -686,6 +734,85 @@ describe('AutoCompoundingRewards', () => {
...
+                                distributionType,
+                                autoCompoundingRewards,
+                                tokens[index].address,
+                                setup.totalRewards,
+                                START_TIME
+                            );
+                        }
+
+                        await autoCompoundingRewards.setAutoProcessRewardsCount(AUTO_PROCESS_REWARDS_COUNT);
+                    });
```
lbeder, line 775: Why is it required?

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -11,7 +11,12 @@ import Contracts, {
     TestPoolCollection,
     TestRewardsMath
 } from '../../components/Contracts';
-import { EXP2_INPUT_TOO_HIGH, RewardsDistributionType, ZERO_ADDRESS } from '../../utils/Constants';
+import {
```
lbeder, line 14: I noticed now that this wasn't added to the `Profile.ts`. We should have a few test cases there as well

lbeder (changes requested):

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -634,7 +639,53 @@ describe('AutoCompoundingRewards', () => {
                     });
                 });
 
-                describe('processing rewards', () => {
+                describe('the number of programs to auto-process the rewards for', () => {
+                    it('should initially be larger than zero', async () => {
+                        expect(await autoCompoundingRewards.autoProcessRewardsCount()).to.be.gt(0);
+                    });
+
+                    it('cannot be set to zero', async () => {
```
lbeder, line 647: All the `it` descriptions don't use our convention. Why is that?

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -686,6 +737,85 @@ describe('AutoCompoundingRewards', () => {
...
+                            for (let i = 0; i < 5; i++) {
+                                await autoCompoundingRewards.setTime(
+                                    Math.floor(START_TIME + (programDuration[distributionType] * 2 ** i) / (2 ** i + 1))
+                                );
+                                const res = await autoCompoundingRewards.autoProcessRewards();
+                                await expect(res).to.emit(autoCompoundingRewards, 'RewardsDistributed');
+                            }
+                        });
+
+                        it('should not redistribute tokens before the minimum time has elapsed', async () => {
```
lbeder, line 798: "redistribute" == "process rewards of some pools"?

test/rewards/AutoCompoundingRewards.ts
lbeder: You shouldn't count 10 events (for example), but rather check for explicit 10 events with correct arguments

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -686,6 +737,85 @@ describe('AutoCompoundingRewards', () => {
...
+                                const res = await autoCompoundingRewards.autoProcessRewards();
+                                await expect(res).to.emit(autoCompoundingRewards, 'RewardsDistributed');
+                            }
+                            const res = await autoCompoundingRewards.autoProcessRewards();
+                            await expect(res).not.to.emit(autoCompoundingRewards, 'RewardsDistributed');
+                        });
+
+                        it('should distribute some tokens', async () => {
+                            for (let i = 0; i < 5; i++) {
+                                await autoCompoundingRewards.setTime(
```
lbeder, line 790: I'm not sure if this achieves what we want if, but in general, it's unnecessary and too complicated (and also, you move the time to the end of all programs at the beginning, while we want to test previous distribution timestamps and also create "holes"). Basically, we need a simple test that:

1. Creates "discontinuous" triggering holes (X V V X X X V)
2. Tests that the hole is being properly skipped over. 

Can you simplify the test and make sure that it's properly tested? You don't need to go crazy and test all possible combinations, but instead, start with a single state (e.g., X V V X X X V), go over it in 1-2 loops (`autoProcessRewards()` cal,s) and make sure that the holes are being filled.

contracts/rewards/AutoCompoundingRewards.sol
barakman: It a product decision, I suppose.
For sure, in terms of user-perspective-efficiency, it means fewer actually-processed programs.
@yudilevi : WDYT?

test/rewards/AutoCompoundingRewards.ts
barakman: This IS the test group.
We are doing various checks on the number of programs to auto-process the rewards for:
- Verifying that it is initially larger than zero
- Verifying that it cannot be set to zero
- Verifying that it cannot be set by a non-admin
- Verifying that it can be set by the admin to a value larger than zero

barakman: BTW, GitHub comparison is a little misleading in this case.
I have not renamed this 'describe' group as this comparison implies.
This is a brand new group added in this PR, right before a 'describe' group which was named 'processing rewards'.
I have renamed that group to 'process rewards', in order for it to be "aligned" with another 'describe' group which I have added in this PR, named 'auto-process rewards'.

test/rewards/AutoCompoundingRewards.ts
barakman: A preliminary initialization before testing function `autoProcessRewards`.

test/rewards/AutoCompoundingRewards.ts
barakman: Because the description - `the number of programs to auto-process the rewards for` - is already given in the 'describe' itself.

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -686,6 +737,85 @@ describe('AutoCompoundingRewards', () => {
...
+                            await autoCompoundingRewards.setTime(programEndTime[distributionType]);
+                            for (let i = 0; i < Math.ceil(setups.length / AUTO_PROCESS_REWARDS_COUNT); i++) {
+                                const res = await autoCompoundingRewards.autoProcessRewards();
+                                await expect(res).to.emit(autoCompoundingRewards, 'RewardsDistributed');
+                            }
+                            const res = await autoCompoundingRewards.autoProcessRewards();
+                            await expect(res).not.to.emit(autoCompoundingRewards, 'RewardsDistributed');
+                        });
+
+                        it('should distribute some tokens', async () => {
```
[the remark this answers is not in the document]
barakman: Yes. If I remember correctly, I borrowed it from the tests for function `processRewards`.

test/rewards/AutoCompoundingRewards.ts
barakman: Yes.
The original description was something like:
```
should not distribute tokens again after less than the minimum time delta for auto-processing the rewards has elapsed
```

contracts/rewards/AutoCompoundingRewards.sol
barakman: done

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -686,6 +737,85 @@ describe('AutoCompoundingRewards', () => {
...
+                            for (let i = 0; i < 5; i++) {
+                                await autoCompoundingRewards.setTime(
+                                    Math.floor(START_TIME + (programDuration[distributionType] * 2 ** i) / (2 ** i + 1))
+                                );
+                                const res = await autoCompoundingRewards.autoProcessRewards();
+                                await expect(res).to.emit(autoCompoundingRewards, 'RewardsDistributed');
```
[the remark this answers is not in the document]
barakman: FYI: my baseline for these tests was the 'process rewards' test group, where event parameters aren't tested.

test/rewards/AutoCompoundingRewards.ts
lbeder: The name of the brand new `describe` should be this either. My comment is relevant for any `describe` - even if it's a sub-group, etc.

test/rewards/AutoCompoundingRewards.ts
lbeder: `autoCompoundingRewards` should be an initialized proxy, so this setting is unneeded, no? Did something break without using it

test/rewards/AutoCompoundingRewards.ts
lbeder: Let's stay aligned with the rest of the code

test/rewards/AutoCompoundingRewards.ts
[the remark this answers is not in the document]
lbeder: Yeah, I remember that you removed the explicit param checks since it was too complicated to do it on time back then and also since it wasn't critical to the test, but in this case - it is. There may be only a single emitted `RewardsDistributed` event with random parameters, and your test assumes, for some reason, that it was emitted 10 times for the correct pool. 

If you want to rely on the event for this test (there are other options, but this one isn't necessarily a bad one either), you can still enumerate and check all the events manually - but only check the first `Token indexed pool` parameter (manually).

test/rewards/AutoCompoundingRewards.ts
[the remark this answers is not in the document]
lbeder: Please update the description in this case

test/rewards/AutoCompoundingRewards.ts
lbeder: So please fix it

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -335,16 +335,12 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
         uint256 maxCount = Math.min(count * 2, numOfPools);
 
         for (uint256 i = 0; i < maxCount; i++) {
-            index = (index + i) % numOfPools;
-            if (_processRewards(Token(_pools.at(index)), true)) {
-                count -= 1;
-                if (count == 0) {
-                    break;
-                }
+            if (_processRewards(Token(_pools.at(index++ % numOfPools)), true) && --count == 0) {
```
lbeder, line 338: Please restore the previous implementation and avoid these shortcuts. They unnecessarily complicate the code and there was nothing wrong with the previous version

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -778,8 +860,45 @@ describe('AutoCompoundingRewards', () => {
                         it('should distribute all tokens', async () => {
                             await autoCompoundingRewards.setTime(programEndTime[distributionType]);
                             for (let i = 0; i < Math.ceil(setups.length / AUTO_PROCESS_REWARDS_COUNT); i++) {
+                                const tokenAmountToDistributeArray: BigNumber[] = new Array<BigNumber>(
+                                    AUTO_PROCESS_REWARDS_COUNT
+                                );
+                                const poolTokenAmountToBurnArray: BigNumber[] = new Array<BigNumber>(
+                                    AUTO_PROCESS_REWARDS_COUNT
+                                );
+                                const remainingRewards: any[] = new Array<BigNumber>(AUTO_PROCESS_REWARDS_COUNT);
```
lbeder, line 869: Don't use `any` if the type is known

test/rewards/AutoCompoundingRewards.ts
```diff
@@ -778,8 +860,45 @@ describe('AutoCompoundingRewards', () => {
...
                                 const res = await autoCompoundingRewards.autoProcessRewards();
-                                await expect(res).to.emit(autoCompoundingRewards, 'RewardsDistributed');
+                                for (let j = 0; j < AUTO_PROCESS_REWARDS_COUNT; j++) {
+                                    const index = AUTO_PROCESS_REWARDS_COUNT * i + j;
+                                    if (index == tokens.length) {
+                                        break;
+                                    }
+                                    await expect(res)
+                                        .to.emit(autoCompoundingRewards, 'RewardsDistributed')
+                                        .withArgs(
```
lbeder, line 895: I know that you did it before my comment, but FYI - you really don't need the exact rewards data to check that a program for TKN1 was processed, while for TKN2 it was not. You only need to check the first parameter.

barakman: Fro. my experimenting, once you add `withArgs`, it expects to find **all** the parameters, though perhaps I did something wrong while experimenting. In any case, it now verifies all of the event parameters.

test/rewards/AutoCompoundingRewards.ts
barakman: already changed.

test/rewards/AutoCompoundingRewards.ts
barakman: Test updated.

test/rewards/AutoCompoundingRewards.ts
lbeder: This is why I keep emphasizing again and again that you should test for this argument **manually** (i.e., without using `to.emit`, but checking `{ events }` directly)

test/rewards/AutoCompoundingRewards.ts
barakman: What's wrong with the current title?

test/rewards/AutoCompoundingRewards.ts
lbeder: Yeah, verifying the rewards data (when it's actually unneeded) makes the tests much more complicated and harder to follow.

test/rewards/AutoCompoundingRewards.ts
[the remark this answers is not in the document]
barakman: Which description, and to what?

test/rewards/AutoCompoundingRewards.ts
barakman: No. The initial value of the counter is 1, and it verified in a different test group.
In this test group, we set the counter to 3, and then we run several tests based on that counter.

test/rewards/AutoCompoundingRewards.ts
barakman: Test updated.

test/rewards/AutoCompoundingRewards.ts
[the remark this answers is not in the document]
lbeder: "should distribute some tokens"

test/rewards/AutoCompoundingRewards.ts
lbeder: Ah, gotcha. I thought it was the default

test/rewards/AutoCompoundingRewards.ts
lbeder: "redistribute". The test should check the triggering - not the actual distribution (not sure why the "re" is there anyway)

contracts/rewards/AutoCompoundingRewards.sol
lbeder: What was the problem?

test/rewards/AutoCompoundingRewards.ts
barakman: done

test/rewards/AutoCompoundingRewards.ts
barakman: done

test/rewards/AutoCompoundingRewards.ts
barakman: updated

test/rewards/AutoCompoundingRewards.ts
barakman: done

lbeder (approved):

contracts/rewards/AutoCompoundingRewards.sol
yudilevi: yup, should return true here (wasn't skipped but rather just didn't have what to distribute - but was fully processed)

contracts/rewards/AutoCompoundingRewards.sol
```diff
@@ -281,26 +304,82 @@ contract AutoCompoundingRewards is IAutoCompoundingRewards, ReentrancyGuardUpgra
...
         uint256 tokenAmountToDistribute = _tokenAmountToDistribute(p, currTime);
         if (tokenAmountToDistribute == 0) {
-            return;
+            return false;
         }
 
         uint256 poolTokenAmountToBurn = _poolTokenAmountToBurn(pool, p, tokenAmountToDistribute);
         if (poolTokenAmountToBurn == 0) {
-            return;
+            return false;
```
yudilevi, line 382: should return `true` here as well

yudilevi (changes requested):

contracts/rewards/AutoCompoundingRewards.sol
barakman: done

contracts/rewards/AutoCompoundingRewards.sol
barakman: done

lbeder: No tests for the change?

barakman: Hmmmmm... I'm not even sure that we have tests for the original, i.e., the tests generally just set up a bunch of programs with some returning false and some returning true (which up until now was returned only on the primary execution path of this function); I'm not sure that we have tests that go into each one of the preliminary 'return' statements.

barakman: Done in the sense that the test covers the following execution paths:
- Return `false` from the 2nd `if` statement (skip due to minimum time period not elapsed)
- Return `true` from the 3rd `if` statement (amount of tokens to distribute is zero)
- Return `true` from the end of the function (tokens distributed and pool tokens burned)

yudilevi (approved):

lbeder (approved):

PR #364, opened 2022-04-23
Add forks setup script `setup:fork`

lbeder:


deployments/setup-fork.ts
```diff
@@ -0,0 +1,210 @@
...
+    console.log(`Funding ${account}...`);
+
+    for (const fundingRequest of fundingRequests) {
+        if (fundingRequest.token === NATIVE_TOKEN_ADDRESS) {
+            await fundingRequest.whale.sendTransaction({
+                value: fundingRequest.amount,
+                to: account
+            });
+
+            continue;
```
barakman, line 72: I'd use `else` instead.

deployments/setup-fork.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    const devAddresses = DEV_ADDRESSES.split(',');
+
+    for (const account of devAddresses) {
+        await fundAccount(account, fundingRequests);
+    }
+
+    console.log();
+
+    const lockDuration = 2;
```
barakman, line 164: I'd put this in the `EnvOptions` object, right next to `FORK_RESEARCH: isResearch`, since they are kind of coupled together anyway.

barakman (dismissed):

deployments/setup-fork.ts
yudilevi: I generally prefer `return`/`continue` to `else` when possible, since it's easier to follow the code (`return`/`continue` -> no need to think about the previous block anymore)

yudilevi (dismissed):
nice :)

deployments/setup-fork.ts
lbeder: It's business logic - not an external configuration

yudilevi (approved):

barakman (approved):

PR #369, opened 2022-04-25
Update coverage report

lbeder:


test/utility/BlockNumber.ts
```diff
@@ -0,0 +1,17 @@
...
+
+describe('BlockNumber', () => {
+    let blockNumber: TestBlockNumber;
+
+    beforeEach(async () => {
+        blockNumber = await Contracts.TestBlockNumber.deploy();
+    });
+
+    it('should return the time of the current block number', async () => {
+        expect(await blockNumber.callStatic.realBlockNumber()).to.equal(
```
barakman, line 13: Why is `callStatic` needed here?

test/utility/Time.ts
```diff
@@ -0,0 +1,15 @@
...
+
+describe('Time', () => {
+    let time: TestTime;
+
+    beforeEach(async () => {
+        time = await Contracts.TestTime.deploy();
+    });
+
+    it('should return the time of the current block', async () => {
+        expect(await time.callStatic.realTime()).to.equal((await ethers.provider.getBlock('latest')).timestamp);
```
barakman, line 13: Why is `callStatic` needed here?

test/utility/BlockNumber.ts
```diff
@@ -0,0 +1,17 @@
...
+            (await ethers.provider.getBlock('latest')).number
```
barakman, line 14: You can use function `latest` from file `BlockNumber.ts`.

test/utility/Time.ts
barakman, line 13: You can use function `latest` from file `Time.ts`.

barakman (changes requested):

test/utility/BlockNumber.ts
lbeder: Because each non-call transaction increments the current block number/time

test/utility/Time.ts
lbeder: ecause each non-call transaction increments the current block number/time

test/utility/BlockNumber.ts
lbeder: 👍

test/utility/Time.ts
lbeder: 👍

yudilevi (approved):

barakman (approved):

PR #373, opened 2022-04-27
Allow the DAO to reduce the trading liquidity of a pool in a safe and controlled manner

barakman:


contracts/pools/PoolCollection.sol
```diff
@@ -207,6 +208,17 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
      */
     event DepositLimitUpdated(Token indexed pool, uint256 prevDepositLimit, uint256 newDepositLimit);
 
+    /**
+     * @dev triggered when a pool's trading liquidity is reduced
+     */
+    event TradingLiquidityReduced(
```
yudilevi, line 214: There's already the TradingLiquidityUpdated event that must be fired when the trading liquidity is updated

contracts/pools/PoolCollection.sol
```diff
@@ -564,6 +576,53 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
...
+        if (bntAmount >= bntTradingLiquidity) {
+            return;
+        }
+
+        if (bntAmount == 0) {
+            _resetTradingLiquidity(bytes32(0), pool, data, TRADING_STATUS_UPDATE_ADMIN);
+            return;
+        }
+
+        if (bntAmount < _networkSettings.poolFundingLimit(pool)) {
```
yudilevi, line 599: would be helpful to have some comments throughout the code to provide the rationale for these conditions, especially this one

contracts/pools/PoolCollection.sol
barakman: So you want to replace the emitting of this event, with emitting the `TradingLiquidityUpdated` event?

barakman: Fixed

contracts/pools/PoolCollection.sol
barakman: Done

PR #381, opened 2022-05-02
Add Return Values to Different BancorNetwork Methods for Simplified Integrations

barakman:
In contract `BancorNetwork`:
- Extend function `cancelWithdrawal` to return the value of the position in units of the reserve token
- Extend function `tradeBySourceAmount/Permitted` to return the trade target amount
- Extend function `tradeByTargetAmount/Permitted` to return the trade source amount
- Extend function `flashLoan` to return the amount of fee paid
- Extend function `withdrawNetworkFees` to return the amount of fees withdrawn

### DIFF
--- a/contracts/network/BancorNetwork.sol
+++ b/contracts/network/BancorNetwork.sol
@@ -836,18 +836,21 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         uint256 amount,
         uint256 availableAmount,
         uint256 originalAmount
-    ) external payable whenNotPaused onlyRoleMember(ROLE_MIGRATION_MANAGER) nonReentrant {
+    ) external payable whenNotPaused onlyRoleMember(ROLE_MIGRATION_MANAGER) nonReentrant returns (uint256) {
         bytes32 contextId = keccak256(
             abi.encodePacked(msg.sender, _time(), token, provider, amount, availableAmount, originalAmount)
         );
 
+        uint256 poolTokenAmount;
         if (token.isEqual(_bnt)) {
-            _depositBNTFor(contextId, provider, amount, msg.sender, true, originalAmount);
+            poolTokenAmount = _depositBNTFor(contextId, provider, amount, msg.sender, true, originalAmount);
         } else {
-            _depositBaseTokenFor(contextId, provider, token, amount, msg.sender, availableAmount);
+            poolTokenAmount = _depositBaseTokenFor(contextId, provider, token, amount, msg.sender, availableAmount);
         }
 
         emit FundsMigrated(contextId, token, provider, amount, availableAmount, originalAmount);
+
+        return poolTokenAmount;
     }
 
     /**
--- a/contracts/network/BancorNetwork.sol
+++ b/contracts/network/BancorNetwork.sol
@@ -839,21 +839,18 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         uint256 amount,
         uint256 availableAmount,
         uint256 originalAmount
-    ) external payable whenNotPaused onlyRoleMember(ROLE_MIGRATION_MANAGER) nonReentrant returns (uint256) {
+    ) external payable whenNotPaused onlyRoleMember(ROLE_MIGRATION_MANAGER) nonReentrant {
         bytes32 contextId = keccak256(
             abi.encodePacked(msg.sender, _time(), token, provider, amount, availableAmount, originalAmount)
         );
 
-        uint256 poolTokenAmount;
         if (token.isEqual(_bnt)) {
-            poolTokenAmount = _depositBNTFor(contextId, provider, amount, msg.sender, true, originalAmount);
+            _depositBNTFor(contextId, provider, amount, msg.sender, true, originalAmount);
         } else {
-            poolTokenAmount = _depositBaseTokenFor(contextId, provider, token, amount, msg.sender, availableAmount);
+            _depositBaseTokenFor(contextId, provider, token, amount, msg.sender, availableAmount);
         }
 
         emit FundsMigrated(contextId, token, provider, amount, availableAmount, originalAmount);
-
-        return poolTokenAmount;
     }
 
     /**
--- a/contracts/network/BancorNetwork.sol
+++ b/contracts/network/BancorNetwork.sol
@@ -784,7 +784,6 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         validAddress(address(recipient))
         whenNotPaused
         nonReentrant
-        returns (uint256)
     {
         if (!token.isEqual(_bnt) && !_networkSettings.isTokenWhitelisted(token)) {
             revert NotWhitelisted();
@@ -826,8 +825,6 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         }
 
         emit FlashLoanCompleted({ token: token, borrower: msg.sender, amount: amount, feeAmount: feeAmount });
-
-        return feeAmount;
     }
 
     /**
--- a/contracts/network/BancorNetwork.sol
+++ b/contracts/network/BancorNetwork.sol
@@ -681,10 +681,10 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         uint256 minReturnAmount,
         uint256 deadline,
         address beneficiary
-    ) external payable whenNotPaused nonReentrant {
+    ) external payable whenNotPaused nonReentrant returns (uint256) {
         _verifyTradeParams(sourceToken, targetToken, sourceAmount, minReturnAmount, deadline);
 
-        _trade(
+        return _trade(
             TradeTokens({ sourceToken: sourceToken, targetToken: targetToken }),
             TradeParams({ bySourceAmount: true, amount: sourceAmount, limit: minReturnAmount }),
             TraderInfo({ trader: msg.sender, beneficiary: beneficiary }),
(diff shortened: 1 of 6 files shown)


contracts/network/interfaces/IBancorNetwork.sol
```diff
@@ -273,25 +279,27 @@ interface IBancorNetwork is IUpgradeable {
         uint256 amount,
         IFlashLoanRecipient recipient,
         bytes calldata data
-    ) external;
+    ) external returns (uint256);
 
     /**
      * @dev deposits liquidity during a migration
+     * returns the amount of pool tokens issued
```
yudilevi, line 286: probably better to use "minted" here as this is what we use all around

contracts/network/interfaces/IPendingWithdrawals.sol
```diff
@@ -66,13 +66,14 @@ interface IPendingWithdrawals is IUpgradeable {
 
     /**
      * @dev cancels a withdrawal request
+     * returns the value of the position in units of the reserve token
```
yudilevi, line 69: `returns the number of reserve tokens associated with the withdrawal request`

yudilevi (commented):
looks good, still missing some tests

contracts/network/interfaces/IBancorNetwork.sol
barakman: done

contracts/network/interfaces/IPendingWithdrawals.sol
barakman: done

test/network/BancorNetwork.ts
```diff
@@ -2117,7 +2117,7 @@ describe('BancorNetwork', () => {
             targetTokenAddress?: string;
         }
 
-        const tradeBySourceAmount = async (amount: BigNumberish, overrides: TradeOverrides = {}) => {
+        const tradeBySourceAmount = async (amount: BigNumberish, overrides: TradeOverrides = {}, getRetVal = false) => {
```
lbeder, line 2120: 1. `getRetVal` (terrible name) can be part of `TradeOverrides` (which can be renamed to `TradeOptions`). Its whole purpose is to group all additional parameters.

2. I don't like the style of passing something of "get ret val" to these functions. It's a strange interface. Let's always return the value in these cases (the performance cost is not that significant, since most of the time is wasted on setup anyway).

@yudilevi other ideas, perhaps?

test/network/BancorNetwork.ts
```diff
@@ -2129,22 +2129,42 @@ describe('BancorNetwork', () => {
...
-                    {
-                        value
-                    }
-                );
+            let retVal = BigNumber.from(0);
+            if (getRetVal) {
+                retVal = await network
+                    .connect(trader)
+                    .callStatic.tradeBySourceAmount(
+                        sourceTokenAddress,
```
lbeder, line 2137: You can group these arguments in an external array and pass is as `...args` in both of the calls.

test/network/BancorNetwork.ts
```diff
@@ -2394,13 +2483,19 @@ describe('BancorNetwork', () => {
...
+                    beneficiary: beneficiaryAddress,
+                    deadline
+                },
+                true
+            );
+
+            expect(retVal).to.equal(hop2.amount);
 
-            const transactionCost = await getTransactionCost(res);
+            const transactionCost = await getTransactionCost(await res);
```
lbeder, line 2498: Let's `await` in `tradeFunc` and then its users won't need to remember to do it

test/network/BancorNetwork.ts
```diff
@@ -3853,6 +3958,10 @@ describe('BancorNetwork', () => {
             });
 
             it('should cancel a pending withdrawal request', async () => {
+                const withdrawalRequest = await pendingWithdrawals.withdrawalRequest(id);
+                const val = await network.connect(provider).callStatic.cancelWithdrawal(id);
```
lbeder, line 3962: Similar to the above, it'd be clearer to give it more specific name

lbeder (commented):
I've left a few suggestions, but neither is critical. Tests also seem to be failing, but I assume that you are already aware.

test/network/BancorNetwork.ts
barakman: That was my initial solution.
However, we also need to be able to call this function when reverting, and then catch the revert and verify the error-message.
Up until now, we'd return the transaction promise, and resolve it by the caller, which would await for either successful completion, or a revert with an error.
Now we need to return two promises - one for the static-call, and the other one for the actual transaction (the same one which we've been returning up until now).
Our options for the first promise (the static-call one) are:
1. Await for (resolve) it before returning the result - that would be a problem, because it may revert.
2. Return it unresolved - that means that we will need to:
   - Make sure to resolve the two promises in the exact same order, in every test which checks both of them (because the 2nd promise changes the state and impact the result of the 1st promise)
   - Resolve each promise also in every test which doesn't make use of that promise (otherwise it's an unresolved promise "left hanging there", I'm not so sure what the consequences are for that); this also applies for tests which verify 'revert with an error', where we will need to expect the same revert on each one of the promises

Hence I chose a 3rd option:
Await for the 1st promise only when explicitly stated by the 'getRetVal' argument, which is set to true only in some specific tests where we're not catching a revert.

test/network/BancorNetwork.ts
barakman: You can't do that, because it sometimes reverts (depending on the test).
That's probably why the author of these tests has returned it unresolved to begin with.

contracts/network/BancorNetwork.sol
```diff
@@ -821,6 +826,8 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         }
 
         emit FlashLoanCompleted({ token: token, borrower: msg.sender, amount: amount, feeAmount: feeAmount });
+
+        return feeAmount;
```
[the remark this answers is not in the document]
barakman: @yudilevi : WDYT?

contracts/network/PendingWithdrawals.sol
```diff
@@ -368,6 +368,8 @@ contract PendingWithdrawals is IPendingWithdrawals, Upgradeable, Time, Utils {
             reserveTokenAmount: request.reserveTokenAmount,
             timeElapsed: _time() - request.createdAt
         });
+
+        return request.reserveTokenAmount;
```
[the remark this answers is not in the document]
barakman: @yudilevi : WDYT? (product)

test/network/BancorNetwork.ts
barakman: That being said, the solution that I've employed is pretty horrific, so I have revised that into something much simpler, namely - each function can return with the transaction itself (as before), or a static-call of it.

test/network/BancorNetwork.ts
barakman: Solution revised

contracts/network/interfaces/IBancorNetwork.sol
```diff
@@ -168,12 +168,13 @@ interface IBancorNetwork is IUpgradeable {
 
     /**
      * @dev cancels a withdrawal request
+     * returns the number of reserve tokens associated with the withdrawal request
```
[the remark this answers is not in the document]
barakman: Fixed (here, and in all other places)

test/network/BancorNetwork.ts
```diff
@@ -2129,22 +2129,42 @@ describe('BancorNetwork', () => {
...
-                    targetTokenAddress,
-                    amount,
-                    minReturnAmount,
-                    deadline,
-                    beneficiary,
-                    {
-                        value
-                    }
-                );
+            let retVal = BigNumber.from(0);
```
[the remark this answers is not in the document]
barakman: Solution revised

test/network/BancorNetwork.ts
barakman: Fixed (in all other places as well)

yudilevi (approved):

test/network/BancorNetwork.ts
```diff
@@ -2117,7 +2117,11 @@ describe('BancorNetwork', () => {
             targetTokenAddress?: string;
         }
 
-        const tradeBySourceAmount = async (amount: BigNumberish, overrides: TradeOverrides = {}, getRetVal = false) => {
+        const tradeBySourceAmount = async (
+            amount: BigNumberish,
+            overrides: TradeOverrides = {},
+            callStatic = false
```
lbeder, line 2123: Why not add this to `TradeOverrides` and rename it to `TradeOptions`?

test/network/BancorNetwork.ts
```diff
@@ -2129,42 +2133,26 @@ describe('BancorNetwork', () => {
...
-                        amount,
-                        minReturnAmount,
-                        deadline,
-                        beneficiary,
-                        {
-                            value
-                        }
-                    );
-            }
+            const caller = callStatic ? network.connect(trader).callStatic : network.connect(trader);
```
lbeder, line 2136: Maybe `caller` is not the best name. I'd rather use something like `method` (since it's basically a method chain)?

contracts/network/BancorNetwork.sol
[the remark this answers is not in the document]
yudilevi: sure, can remove

contracts/network/PendingWithdrawals.sol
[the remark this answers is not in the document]
yudilevi: makes sense

contracts/network/BancorNetwork.sol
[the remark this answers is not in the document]
lbeder: Done

contracts/network/PendingWithdrawals.sol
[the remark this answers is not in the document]
lbeder: Done

contracts/network/interfaces/IBancorNetwork.sol
```diff
@@ -167,8 +167,7 @@ interface IBancorNetwork is IUpgradeable {
     ) external returns (uint256);
 
     /**
-     * @dev cancels a withdrawal request
-     * returns the number of reserve tokens associated with the withdrawal request
+     * @dev cancels a withdrawal request, and returns the number of reserve tokens associated with the withdrawal request
```
yudilevi, line 170: it's now changed to return the pool token amount associated with the request

lbeder (approved):

yudilevi (approved):

PR #385, opened 2022-05-04
Add a system snapshot taking script

lbeder:


yudilevi (approved):

scripts/snapshot.ts
```diff
@@ -0,0 +1,163 @@
...
+
+const gcd = (x: BigNumber, y: BigNumber) => {
+    while (y.gt(0)) {
+        const t = y;
+        y = x.mod(y);
+        x = t;
+    }
+
+    return x;
+};
```
barakman, line 56: As far as i can tell, that does not handle correctly cases where `x < y`.
You can amend it via `[x, y] = x.gt(y) ? [x, y] : [y, x]`.

scripts/snapshot.ts
```diff
@@ -0,0 +1,163 @@
...
+
+const saveCSV = (data: Record<string, any>[], outputPath: string) => {
+    const headers = Object.keys(data[0]);
+    const rows = data
+        .map((p) =>
+            Object.values(p)
+                .map((v) => v.toString())
```
barakman, line 63: Using `toString()` on `Decimal` values may yield scientific-notation strings.
In order to avoid that, you can use `toFixed()` specifically for `Decimal` values.

scripts/snapshot.ts
```diff
@@ -0,0 +1,163 @@
...
+    await saveGlobalStats();
+    await savePoolStats();
+};
+
+main()
+    .then(() => process.exit(0))
+    .catch((error) => {
+        console.error(error);
+        process.exit(1);
+    });
```
barakman, line 163: You can just do `main()` at the very bottom of the script as far as I know (though perhaps there is some advantage in this code which I am not aware of).

barakman (changes requested):

scripts/snapshot.ts
lbeder: It's a more structure way to handle and output the errors, but a plain `main` would've worked as well (with our version of node)

scripts/snapshot.ts
lbeder: I think that if `x < y`, then the first loop will already switch between them as well.

scripts/snapshot.ts
lbeder: It should be ok for this case, since this CSV is being used in as a spreadsheet

scripts/snapshot.ts
barakman: That's right, I didn't stop to think about it, sorry :(

scripts/snapshot.ts
barakman: The problem is that once converted to scientific-notation, you lose data (i.e., the accuracy of the stored information is degraded).

scripts/snapshot.ts
```diff
@@ -0,0 +1,163 @@
...
+    const { dai, link } = await getNamedAccounts();
+
+    const pools = {
+        ETH: NATIVE_TOKEN_ADDRESS,
+        LINK: link,
+        DAI: dai
+    };
+
+    const settings = await DeployedContracts.NetworkSettings.deployed();
+    const poolCollection = await DeployedContracts.PoolCollectionType1V1.deployed();
```
yudilevi, line 106: should be v2 now

yudilevi (changes requested):

lbeder:
Closing as this is no longer required

PR #387, opened 2022-05-06
 Switch CI to Tenderly + Add more tests and Improve the Pool Stabilization Method + Upgrade the BancorNetwork and the PendingWithdrawal Contracts

lbeder:


deploy/tests/_setup.ts
```diff
@@ -0,0 +1,51 @@
...
+import { HttpNetworkUserConfig } from 'hardhat/types';
+
+interface EnvOptions {
+    TENDERLY_USERNAME: string;
+    TENDERLY_ACCESS_KEY: string;
+}
+
+const { TENDERLY_USERNAME, TENDERLY_ACCESS_KEY }: EnvOptions = process.env as any as EnvOptions;
+
+const TENDERLY_TEST_PROJECT_NAME = 'v3-temp-forks';
```
yudilevi, line 13: shouldn't this be taken from the .env file?

lbeder: The rationale was to use a hard-coded "system project" that no one should configure, but I guess we can also move it to `.env` and specify it explicitly in workflow files

lbeder: Fixed

yudilevi (approved):

yudilevi (approved):

PR #396, opened 2022-05-11
Enable no-floating-promises and prefer-nullish-coalescing TypeScript linter rules

lbeder:
- `no-floating-promises`: prevents unresolved promises (such as forgetting an `await` or not returning a promise from an `async` function
- `prefer-nullish-coalescing`: prefer chaining parameters using `??` instead of `||` (so that both of the vars checked for type compatibility)


yudilevi (approved):
nice :)

PR #397, opened 2022-05-13
Update dependencies

lbeder:


yudilevi (approved):

lbeder:
There are some backward compatibility issue that I missed, so stay tuned

PR #406, opened 2022-05-19
PoolCollection V3

lbeder:
- Added new query functions to the  new `PoolCollection` and `BancorNetworkInfo` contracts
- Removed deposit limits 
- Updated `PoolMigrator` accordingly
- Simplified the calculation of withdrawal pool token and reserve token amounts
- Improve pool management events
- Support inverse EMA
- Update the EMAs upon deposits and withdrawals
- Improved the tests
- Refactoring


contracts/network/BancorNetworkInfo.sol
```diff
@@ -242,6 +242,34 @@ contract BancorNetworkInfo is IBancorNetworkInfo, Upgradeable, Utils {
...
+     */
+    function tradingFeePPM(Token pool) external view returns (uint32) {
+        return pool.isEqual(_bnt) ? 0 : _poolCollection(pool).tradingFeePPM(pool);
+    }
+
+    /**
+     * @inheritdoc IBancorNetworkInfo
+     */
+    function tradingEnabled(Token pool) external view returns (bool) {
+        return pool.isEqual(_bnt) ? true : _poolCollection(pool).tradingEnabled(pool);
```
barakman, line 256: Minor (for this change and for the two changes which follow it):
```
return pool.isEqual(_bnt) || _poolCollection(pool).tradingEnabled(pool);
```

contracts/pools/PoolCollection.sol
```diff
@@ -848,21 +891,15 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
     function _poolWithdrawalAmounts(
         Token pool,
         Pool memory data,
-        uint256 poolTokenAmount
+        uint256 poolTokenAmount,
+        uint256 reserveTokenAmount,
```
barakman, line 895: Probably better to use the original variable name here, i.e., `baseTokensWithdrawalAmount` instead of `reserveTokenAmount`, or by the least, something which starts with 'baseToken', in order to avoid having "3 different types of tokens" appearing in this context (pool, base, reserve).

contracts/pools/PoolCollection.sol
```diff
@@ -647,7 +675,22 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
         greaterThanZero(poolTokenAmount)
         returns (WithdrawalAmounts memory)
     {
-        InternalWithdrawalAmounts memory amounts = _poolWithdrawalAmounts(pool, _poolStorage(pool), poolTokenAmount);
+        Pool storage data = _poolData[pool];
```
barakman, line 678: Some fields in `data.liquidity` (like `stakedBalance` and `baseTokenTradingLiquidity`) are being read from storage several times in this context (i.e., in this function and in the function called from this function).

Probably better to read `data.liquidity` into memory once, then use it in this function and in the other function (especially because that other function takes it by `memory`, which means that the entire `data` object is read from storage into memory before that function is called anyway).

In fact, that other function requires only `data.liquidity` and `data. tradingFeePPM` from that object, so probably better to just pass those two fields.

contracts/pools/PoolCollection.sol
lbeder: Agreed 👍

contracts/helpers/TestPoolCollection.sol
```diff
@@ -62,12 +62,23 @@ contract TestPoolCollection is PoolCollection, TestBlockNumber {
...
+    ) external view returns (InternalWithdrawalAmounts memory) {
+        Pool storage data = _poolData[pool];
+        PoolLiquidity memory liquidity = data.liquidity;
+
+        return
+            _poolWithdrawalAmounts(
+                pool,
+                poolTokenAmount,
+                baseTokensWithdrawalAmount,
+                liquidity,
```
barakman, line 78: Semantic: you can just put `data.liquidity` here directly.

contracts/pools/PoolCollection.sol
```diff
@@ -624,15 +627,42 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
         bytes32 contextId,
         address provider,
         Token pool,
-        uint256 poolTokenAmount
-    ) external only(address(_network)) validAddress(provider) greaterThanZero(poolTokenAmount) returns (uint256) {
+        uint256 poolTokenAmount,
+        uint256 reserveTokenAmount
```
barakman, line 631: Perhaps `baseTokenAmount` here too?

contracts/pools/PoolCollection.sol
```diff
@@ -624,15 +627,42 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
...
+        if (reserveTokenAmount > underlyingAmount) {
+            revert InvalidParam();
+        }
 
         // obtain the withdrawal amounts
-        InternalWithdrawalAmounts memory amounts = _poolWithdrawalAmounts(pool, data, poolTokenAmount);
+        InternalWithdrawalAmounts memory amounts = _poolWithdrawalAmounts(
+            pool,
+            poolTokenAmount,
+            reserveTokenAmount,
```
barakman, line 658: If I understand correctly, we previously used the entire `underlyingAmount` here.
Now we are using the possibly smaller `reserveTokenAmount` instead.
Shouldn't the diff `underlyingAmount - reserveTokenAmount` be dealt with?

lbeder: We pass the exact TKN that was calculated during the initiation of the withdrawal request

contracts/helpers/TestPoolCollection.sol
lbeder: 👍

contracts/pools/PoolCollection.sol
lbeder: Sure

contracts/network/BancorNetwork.sol
```diff
@@ -665,10 +665,10 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         CompletedWithdrawal memory completedRequest = _pendingWithdrawals.completeWithdrawal(contextId, provider, id);
 
         if (completedRequest.poolToken == _bntPoolToken) {
-            return _withdrawBNT(contextId, provider, completedRequest);
+            return _withdrawBNT(contextId, provider, _pendingWithdrawals, completedRequest);
```
yudilevi, line 668: why pass a state var to an internal method?

contracts/pools/PoolCollection.sol
```diff
@@ -884,13 +925,14 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
                 bntProtocolHoldingsDelta: output.q,
                 baseTokensWithdrawalFee: output.v,
                 baseTokensWithdrawalAmount: baseTokensWithdrawalAmount,
+                poolTokenAmount: poolTokenAmount,
```
yudilevi, line 928: I was trying to avoid involving the pool tokens in withdrawal amount calculations - why do we need it here?

contracts/network/BancorNetwork.sol
lbeder: Since it's not immutable, I wanted to save the extra read. Too much?

contracts/pools/PoolCollection.sol
lbeder: OCD, mostly, since we are already passing almost everything else using this struct (due to `stack too deep` issues) and this amount needs to be passed down anyway

contracts/network/BancorNetwork.sol
yudilevi: I must be missing something - where is the extra read?

contracts/pools/PoolCollection.sol
yudilevi: ok, makes sense, as long as we don't start adding it to `calculateWithdrawalAmounts`

contracts/network/BancorNetwork.sol
barakman: Two lines above, but it wasn't cached obviously

lbeder: Discussed and removed 👍

yudilevi (dismissed):

yudilevi (approved):

contracts/pools/PoolMigrator.sol
```diff
@@ -136,25 +138,31 @@ contract PoolMigrator is IPoolMigrator, Upgradeable, Utils {
...
             poolToken: data.poolToken,
             tradingFeePPM: data.tradingFeePPM,
             tradingEnabled: data.tradingEnabled,
             depositingEnabled: data.depositingEnabled,
-            averageRate: data.averageRate,
-            depositLimit: data.depositLimit,
+            averageRates: AverageRates({
+                blockNumber: averageRate.blockNumber,
+                rate: averageRate.rate,
+                invRate: Fraction({ n: liquidity.baseTokenTradingLiquidity, d: liquidity.bntTradingLiquidity })
```
barakman, line 159: Shouldn't we take the inverse of `averageRate.rate` upon migration?

test/network/BancorNetwork.ts
```diff
@@ -3910,6 +3926,8 @@ describe('BancorNetwork Financial Verification', () => {
         bntTradingLiquidity: Decimal;
         averageRateN: Decimal;
         averageRateD: Decimal;
+        invAverageRateN: Decimal;
+        invAverageRateD: Decimal;
```
barakman, line 3930: `averageInvRate` would probably be more accurate (otherwise, it's just the inverse of the original average rate)

test/network/BancorNetwork.ts
```diff
@@ -4071,8 +4099,10 @@ describe('BancorNetwork Financial Verification', () => {
         actual.bntStakedBalance = integerToDecimal(await bntPool.stakedBalance(), bntDecimals);
         actual.tknTradingLiquidity = integerToDecimal(poolData.liquidity.baseTokenTradingLiquidity, tknDecimals);
         actual.bntTradingLiquidity = integerToDecimal(poolData.liquidity.bntTradingLiquidity, bntDecimals);
-        actual.averageRateN = integerToDecimal(poolData.averageRate.rate.n, 0);
-        actual.averageRateD = integerToDecimal(poolData.averageRate.rate.d, 0);
+        actual.averageRateN = integerToDecimal(poolData.averageRates.rate.n, 0);
+        actual.averageRateD = integerToDecimal(poolData.averageRates.rate.d, 0);
+        actual.invAverageRateN = integerToDecimal(poolData.averageRates.rate.n, 0);
+        actual.invAverageRateD = integerToDecimal(poolData.averageRates.rate.d, 0);
```
barakman, line 4105: Should use `invRate` here

barakman (changes requested):

contracts/pools/PoolCollection.sol
```diff
@@ -872,6 +889,56 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
...
+
+        // if the requested trading liquidity is smaller than the funding limit - revert (the funding limit should be
+        // decreased below the new value first)
+        if (newTradingLiquidity < _networkSettings.poolFundingLimit(pool)) {
+            revert FundingLimitTooHigh();
+        }
+
+        // deplete the extra amount of funding currently available
+        uint256 renouncedAmount = bntTradingLiquidity - newTradingLiquidity;
+        _bntPool.renounceFunding(contextId, pool, renouncedAmount);
```
yudilevi, line 926: this will reduce the current funding, while we want to reduce the trading liquidity and deplete the available funding - we actually need to request the entire available funding and burn the BNT, which will result in the available funding being equal to 0. The idea is that no further funding can be made after the reduction

contracts/helpers/TestFractionLibrary.sol
```diff
@@ -26,18 +26,26 @@ contract TestFractionLibrary {
...
+
     function isPositive112(Fraction112 memory fraction) external pure returns (bool) {
         return fraction.isPositive();
     }
 
+    function inverse256(Fraction memory fraction) external pure returns (Fraction memory) {
+        return fraction.inverse();
+    }
+
+    function inverse112(Fraction memory fraction) external pure returns (Fraction memory) {
```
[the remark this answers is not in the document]
lbeder: Yep 👍

contracts/pools/PoolCollection.sol
```diff
@@ -1497,48 +1612,64 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
...
 
-        if (MathEx.isInRange(averageRate.fromFraction112(), spotRate, RATE_MAX_DEVIATION_PPM)) {
+        if (
+            MathEx.isInRange(rate.fromFraction112(), spotRate, RATE_MAX_DEVIATION_PPM) &&
+            MathEx.isInRange(invRate.fromFraction112(), invSpotRate, RATE_MAX_DEVIATION_PPM)
+        ) {
             return PoolRateState.Stable;
         }
 
         return PoolRateState.Unstable;
```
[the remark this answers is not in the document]
lbeder: There is some duplication and some nuances (for example, you can't initialize the inverse if the spot rate has a `0` denominator and should check the block number for both), so I think that over refactoring will ugflify the code. Let's get back to this at the end.

contracts/pools/PoolMigrator.sol
lbeder: Since we don't actually know the aggregated value of the inverse, I'd figured to start with it to be exactly equal to the rate and let it diverge from there. @yudilevi WDYT?

test/network/BancorNetwork.ts
lbeder: 👍

test/network/BancorNetwork.ts
lbeder: Right 👍

yudilevi (changes requested):

contracts/pools/PoolCollection.sol
```diff
@@ -1497,48 +1612,64 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
...
+                invRate: _calcAverageRate(data.averageRates.invRate, spotRate.inverse())
             });
         }
     }
 
     /**
      * @dev calculates the average rate
      */
-    function _calcAverageRate(Fraction112 memory averageRate, Fraction memory spotRate)
+    function _calcAverageRate(Fraction112 memory averageRate, Fraction memory poolRate)
```
barakman, line 1661: Semantic: I don't think that we should use `poolRate` here when we use `spotRate` and `SPOT_RATE` everywhere else.

contracts/pools/PoolCollection.sol
```diff
@@ -594,10 +603,18 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
...
-            data.averageRate.rate.fromFraction112(),
+            data.averageRates.rate.fromFraction112(),
             _networkSettings.minLiquidityForTrading()
         );
 
+        // if trading is enabled, then update the recent average rates
+        if (data.tradingEnabled) {
+            _updateAverageRates(
+                data,
+                Fraction({ n: data.liquidity.bntTradingLiquidity, d: data.liquidity.baseTokenTradingLiquidity })
```
barakman, line 614: By now, we probably need a pure getter function to take `PoolLiquidity` as input and return the spot-rate as output (can be used in a few places, including the one where we comprise the inverse spot-rate).

contracts/pools/PoolCollection.sol
lbeder: `spotRate` is a bad name because it can be both the spot and the inverse spot rate. I'll change it to `rate`, but in general - please refrain from such nitpicking on tightly scheduled PRs.

contracts/pools/PoolCollection.sol
lbeder: I'll open a task. Closing for now.

yudilevi (approved):

contracts/network/BancorNetwork.sol
```diff
@@ -167,10 +167,20 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
     );
 
     /**
-     * @dev triggered when a new pool is added
+     * @dev triggered when a pool is created
+     */
+    event PoolCreated(Token indexed pool, IPoolCollection indexed poolCollection);
```
yudilevi, line 172: I think we don’t even need to include the pool collection here since the association is available in the PoolAdded event. WDYT?

contracts/network/BancorNetwork.sol
```diff
@@ -167,10 +167,20 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
...
+
+    /**
+     * @dev triggered when a new pool is added to a pool collection
      */
     event PoolAdded(Token indexed pool, IPoolCollection indexed poolCollection);
 
+    /**
+     * @dev triggered when a new pool is removed to a pool collection
```
yudilevi, line 180: Type

contracts/network/BancorNetwork.sol
```diff
@@ -510,7 +520,7 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         // add the pool collection to the reverse pool collection lookup
         _collectionByPool[token] = poolCollection;
 
-        emit PoolAdded({ pool: token, poolCollection: poolCollection });
+        emit PoolCreated({ pool: token, poolCollection: poolCollection });
```
yudilevi, line 523: PoolAdded should be triggered here as well

contracts/network/BancorNetwork.sol
```diff
@@ -523,12 +533,13 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
...
-            if (newPoolCollection == IPoolCollection(address(0))) {
-                continue;
-            }
+            IPoolCollection prevPoolCollection = _collectionByPool[pool];
 
             // update the mapping between pools and their respective pool collections
             _collectionByPool[pool] = newPoolCollection;
+
+            emit PoolAdded(pool, newPoolCollection);
+            emit PoolRemoved(pool, prevPoolCollection);
```
yudilevi, line 542: Removed should be triggered before Added

yudilevi (changes requested):

contracts/network/BancorNetwork.sol
lbeder: Let's keep it in symmetry with the rest of the `PoolXXX` events. We don't know how they are going to be used and if in tandem or not (or not at all)

contracts/network/BancorNetwork.sol
lbeder: 👍

contracts/network/BancorNetwork.sol
lbeder: 👍

contracts/network/BancorNetwork.sol
lbeder: 👍

lbeder: 👍

contracts/network/BancorNetwork.sol
yudilevi: sounds good

yudilevi (approved):

contracts/pools/PoolCollection.sol
```diff
@@ -970,7 +1027,8 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
 
         // ensure that the average rate is reset when the pool is being emptied
         if (amounts.newBaseTokenTradingLiquidity == 0) {
-            data.averageRate.rate = zeroFraction112();
+            data.averageRates.rate = zeroFraction112();
+            data.averageRates.invRate = zeroFraction112();
```
barakman, line 1031: At least mathematically, this initialization (occurring in a few other places) makes no sense, because the inverse of 0 is infinity. I am not entirely convinced that this is just a philosophical debate
We should probably think of a different initialization.
One option is to let both rates represent zero liquidity (i.e., set each rate to the invalid fraction 0:0).
Another option is to let each rate represent a liquidity of 1 BNT and 1 TKN.
So for example, if TKN is USDC, then:
- Set `rate = {n: 1000000000000000000, d: 1000000}`
- Set `rate = {n: 1000000, d: 1000000000000000000}`

The problem with the second option is that we need to rely on `decimals`.

lbeder: This is only an initial value and is being overridden after `enableTrading`, so this comment is irrelevant.

contracts/pools/PoolCollection.sol
```diff
@@ -350,19 +342,15 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
             tradingFeePPM: _defaultTradingFeePPM,
             tradingEnabled: false,
             depositingEnabled: true,
-            averageRate: AverageRate({ blockNumber: 0, rate: zeroFraction112() }),
-            depositLimit: 0,
+            averageRates: AverageRates({ blockNumber: 0, rate: zeroFraction112(), invRate: zeroFraction112() }),
```
barakman, line 345: At present, function `depositFor` calls function `_updateTradingLiquidity` with `rate`.
A possible issue is when `rate` and `invRate` are far apart from each other (i.e., not close to being inverses of each other).
Depending on which one of them is used, the outcome may be very different.
The impact of this is mostly on the output trading liquidities (of BNT and of TKN).

lbeder: It's not different than before. The protocol allows depositing liquidity at any rate (even when there is no rate during the bootstrapping) phase. Still, it won't be possible to withdraw as long as the pool is unstable. Noted.

PR #427, opened 2022-06-12
Prevent accidental deposit of ETH when depositing BNT

lbeder:


contracts/network/BancorNetwork.sol
```diff
@@ -902,6 +902,10 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         bool isMigrating,
         uint256 originalAmount
     ) private returns (uint256) {
+        if (msg.value > 0) {
```
yudilevi, line 905: better not access msg.sender in a non external method, can have side effects if the logic changes.

yudilevi (changes requested):

contracts/network/BancorNetwork.sol
yudilevi: nm, confused with msg.sender

yudilevi (approved):

yudilevi (approved):

yudilevi (approved):

PR #444, opened 2022-07-05
Added PoolCollection v8 & increased the network fee to 30%

yudilevi:


lbeder (commented):
Please note that the tests are failing, but otherwise it looks good

PR #446, opened 2022-07-06
PoolCollection - owner only updateTradingLiquidity method

yudilevi:


test/pools/PoolCollection.ts
```diff
@@ -1028,6 +1028,613 @@ describe('PoolCollection', () => {
         }
     });
 
+    describe('update trading liquidity', () => {
+        const testDeposit = (tokenData: TokenData) => {
```
lbeder, line 1032: `testUpdate`?

test/pools/PoolCollection.ts
```diff
@@ -1028,6 +1028,613 @@ describe('PoolCollection', () => {
...
+                    await networkSettings.setFundingLimit(token.address, MAX_UINT256);
+                    await poolCollection.setBlockNumber(await latestBlockNumber());
+                });
+
+                enum TradingLiquidityState {
+                    Reset = 0,
+                    Ignore = 1,
+                    Update = 2,
+                    InvalidState = 3
+                }
```
lbeder, line 1096: This enum (and some of the code below) can be extracted to the global scope and shared between the different cases

test/pools/PoolCollection.ts
yudilevi: Oops, fixed.

PR #456, opened 2022-07-27
Pool collection fixes

yudilevi:


contracts/pools/PoolCollection.sol
```diff
@@ -1726,6 +1743,11 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
         view
         returns (AverageRates memory)
     {
+        // if the spot rate is 0, reset the average rates
+        if (!spotRate.isPositive()) {
+            return AverageRates({ blockNumber: 0, rate: zeroFraction112(), invRate: zeroFraction112() });
+        }
```
barakman, line 1749: Note that this will not really reset the average rate (and its block number), but rather, just return a "zerofied" average rate.
So if the intention is to actually get the `AverageRates` structure in storage reset to zero, then we should probably apply it separately in each one of the 4 places where this function is called (in one of them, it will not even be called when the spot rate is 0).

barakman (changes requested):

contracts/pools/PoolCollection.sol
```diff
@@ -667,24 +665,24 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
...
         }
 
-        PoolLiquidity memory prevLiquidity = data.liquidity;
-        uint256 currentStakedBalance = prevLiquidity.stakedBalance;
-
         // if there are no pool tokens available to support the staked balance - reset the
         // trading liquidity and the staked balance
         // in addition, get the effective average rates
         uint256 prevPoolTokenTotalSupply = data.poolToken.totalSupply();
+        uint256 currentStakedBalance = data.liquidity.stakedBalance;
```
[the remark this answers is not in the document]
yudilevi: I don't get it - it's used 2 lines later, before the `else` clause

contracts/pools/PoolCollection.sol
```diff
@@ -667,24 +665,24 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
             revert DepositingDisabled();
         }
 
-        PoolLiquidity memory prevLiquidity = data.liquidity;
-        uint256 currentStakedBalance = prevLiquidity.stakedBalance;
```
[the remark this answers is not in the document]
yudilevi: yeah it's there because of a `stack too deep` error

contracts/pools/PoolCollection.sol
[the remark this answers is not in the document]
barakman: It is set to `0` at the beginning of the `if`, so you may as well just read it from storage at the beginning of the `else`.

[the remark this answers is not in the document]
barakman: My mistake; abort.

contracts/pools/PoolCollection.sol
```diff
@@ -715,28 +730,13 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
 
         // if trading is enabled, then update the recent average rates
         if (data.tradingEnabled) {
+            PoolLiquidity memory liquidity = data.liquidity;
```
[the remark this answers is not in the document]
yudilevi: yup, confirmed it should use the updated (post withdrawal) liquidity

contracts/pools/PoolCollection.sol
yudilevi: Correct - the method is supposed to return the "calculated" average rates.
When a storage update is needed, the `_updateAverageRates` method is used and it's called during all relevant places - trade, deposit and withdrawal

contracts/pools/PoolCollection.sol
```diff
@@ -421,8 +421,9 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
      *
      * notes:
      *
-     * - there is no guarantee that this function will remains forward compatible, so please avoid relying on it and
-     *   rely on specific getters from the IPoolCollection interface instead
+     * - there is no guarantee that this function will remains forward compatible,
```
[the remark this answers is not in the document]
yudilevi: fixed

contracts/pools/PoolCollection.sol
```diff
@@ -1250,66 +1260,91 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
...
+        // liquidity increase
+        // note that liquidity increase is capped
+        if (targetBNTTradingLiquidity > liquidity.bntTradingLiquidity) {
+            if (liquidity.bntTradingLiquidity == 0) {
+                // the current BNT trading liquidity is 0 - set it to the minimum liquidity for trading (with an
+                // additional buffer so that initial trades will be less likely to trigger disabling of trading)
+                uint256 newTargetBNTTradingLiquidity = minLiquidityForTrading * BOOTSTRAPPING_LIQUIDITY_BUFFER_FACTOR;
+
+                // ensure that the new target isn't higher than the previously established target
+                targetBNTTradingLiquidity = Math.min(targetBNTTradingLiquidity, newTargetBNTTradingLiquidity);
```
barakman, line 1318: At least according to the comment, this statement should be:
```
newTargetBNTTradingLiquidity = Math.min(targetBNTTradingLiquidity, newTargetBNTTradingLiquidity);
```

barakman: OK, I've confirmed that this confusion stems only from the name `newTargetBNTTradingLiquidity`.

I would just use a single assignment line in this 'if' clause:
```
targetBNTTradingLiquidity = Math.min(
    targetBNTTradingLiquidity,
    minLiquidityForTrading * BOOTSTRAPPING_LIQUIDITY_BUFFER_FACTOR
);
```

yudilevi: Cleared up the docs/name

yudilevi: Nice, updated.

barakman: Or even merge the 'if' and the 'else' as follows:
```
targetBNTTradingLiquidity = Math.min(
    targetBNTTradingLiquidity,
    minLiquidityForTrading * (
        liquidity.bntTradingLiquidity == 0
        ? BOOTSTRAPPING_LIQUIDITY_BUFFER_FACTOR
        : LIQUIDITY_GROWTH_FACTOR
    )
);
```

yudilevi: That I think will make it less readable, I prefer to keep the two cases separate

yudilevi: Actually making another change there

yudilevi: Ok, take a look

contracts/pools/PoolCollection.sol
```diff
@@ -1250,66 +1260,91 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
...
+        baseTokenTradingLiquidityDelta = MathEx.mulDivF(bntTradingLiquidityDelta, fundingRate.d, fundingRate.n);
 
         return
-            TradingLiquidityAction({
+            TargetTradingLiquidity({
                 update: true,
-                newBNTTradingLiquidity: targetBNTTradingLiquidity,
-                newBaseTokenTradingLiquidity: liquidity.baseTokenTradingLiquidity + baseTokenTradingLiquidityDelta
+                bnt: targetBNTTradingLiquidity,
+                baseToken: liquidity.baseTokenTradingLiquidity - baseTokenTradingLiquidityDelta
```
barakman, line 1348: I mas not convinced that the condition `liquidity.baseTokenTradingLiquidity >= baseTokenTradingLiquidityDelta` is mathematically guaranteed at this point (it might, but I cannot easily infer that from the code).
So as a safety measure, we might be better off with:
```
baseToken: MathEx.subMax0(liquidity.baseTokenTradingLiquidity, baseTokenTradingLiquidityDelta)
```
So if `liquidity.baseTokenTradingLiquidity < baseTokenTradingLiquidityDelta` under any circumstances, then the code above will not revert the transaction, and the resulting `baseToken: 0` will lead the calling function to reset the trading liquidity.

yudilevi: Makes sense - updated.

contracts/pools/PoolCollection.sol
barakman: Looks good!

barakman (approved):

PR #463, opened 2022-11-15
Update deprecated commands in Github Actions

adamshai:
- Update node12 to node16
- Replace command `::set-output` that stores env variables between steps, with redirect to `$GITHUB_OUTPUT`


adamshai:
[v3 - fix nightly test](https://trello.com/c/7Gxy5y16/1251-v3-fix-nightly-test)

PR #470, opened 2023-02-28
Arb contract

yudilevi:


contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+    Rewards internal _rewards;
+
+    // upgrade forward-compatibility storage gap
+    uint256[MAX_GAP - 2] private __gap;
+
+    /**
+     * @dev triggered after a successful arb is executed
+     */
+    event ArbitrageExecuted(
+        address caller,
```
barakman, line 106: Probably better declared `indexed`.

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+        uint256 minTargetAmount;
+        uint256 deadline;
+        address customAddress;
+        uint256 customInt;
+    }
+
+    // rewards settings
+    struct Rewards {
+        uint32 percentagePPM;
+        uint256 maxAmount;
```
barakman, line 62: Can probably use `uint224 maxAmount` and reduce this to a single slot.

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+     */
+    function rewards() external view returns (Rewards memory) {
+        return _rewards;
+    }
+
+    /**
+     * @dev execute multi-step arbitrage trade between exchanges
+     */
+    function execute(
+        Route[] memory routes,
```
barakman, line 251: Can use `calldata` here, as well as in each one of the following places:
- `validRouteLength`
- `_validRouteLength`
- `allocateRewards`

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+        if ((address(routes[routes.length - 1].targetToken) != address(_bnt))) {
+            revert InvalidInitialAndFinalTokens();
+        }
+
+        // take a flashloan for the source amount on Bancor v3 and perform the trades
+        _bancorNetworkV3.flashLoan(
+            Token(address(_bnt)),
+            sourceAmount,
+            IFlashLoanRecipient(address(this)),
+            abi.encode(routes, sourceAmount)
```
barakman, line 264: Any security issue with having a token appearing in `routes` more than once?

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+                minTargetAmount,
+                deadline,
+                address(0x0)
+            );
+
+            return;
+        }
+
+        if (exchangeId == EXCHANGE_ID_UNISWAP_V2 || exchangeId == EXCHANGE_ID_SUSHISWAP) {
+            IUniswapV2Router02 router = exchangeId == EXCHANGE_ID_UNISWAP_V2 ? _uniswapV2Router : _sushiSwapRouter;
```
barakman, line 374: Can probably store the immutable addresses of `_uniswapV2Router.WETH()` and `_sushiSwapRouter.WETH()` upon contract creation, then use them here instead of the external call taking place on every trade.
It would require splitting the code below into two separate cases, one for each router, so probably best to use a common function in order to avoid coding-duplication.

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+            }
+
+            // allow the router to withdraw the source tokens
+            Token(tokenIn).safeApprove(address(_uniswapV3Router), sourceAmount);
+
+            // build the params
+            ISwapRouter.ExactInputSingleParams memory params = ISwapRouter.ExactInputSingleParams({
+                tokenIn: tokenIn,
+                tokenOut: tokenOut,
+                fee: uint24(customInt), // fee
```
barakman, line 415: This is a dangerous (unchecked) cast.
Better use OZ's `SafeCast.toUint24()`.

barakman (changes requested):

contracts/arbitrage/BancorArbitrage.sol
yudilevi: good catch - added

contracts/arbitrage/BancorArbitrage.sol
yudilevi: I prefer to keep it at default

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+     * @dev performs contract-specific initialization
+     */
+    function __BancorArbitrage_init_unchained() internal onlyInitializing {
+        _rewards = Rewards({ percentagePPM: 100000, maxAmount: 100 * 1e18 });
+    }
+
+    /**
+     * @dev authorize the contract to receive the native token
+     */
+    receive() external payable {}
```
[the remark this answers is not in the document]
yudilevi: yup - the contract receives ETH during trades on other exchanges

contracts/arbitrage/BancorArbitrage.sol
yudilevi: done

contracts/arbitrage/BancorArbitrage.sol
yudilevi: not really - some trades might be through the same token, in the worst case the arb will be sub-optimal but I don't quite see any security issue as each trade is isolated.

contracts/arbitrage/BancorArbitrage.sol
yudilevi: already known to be the same WETH for both of them on mainnet so can simply use the already defined state var for that - done.

contracts/arbitrage/BancorArbitrage.sol
yudilevi: toUint24 isn't supported on SafeCast, but in this case it's not really an issue - an overflow would mean that the function won't find the pool with the specified fee and the tx will revert. and if it does find a pool (practically impossible), it would use a pool with different fee for a sub-optimal arb which will either revert later on (can't return flashloan) or will be a bit less profitable, so no harm done

contracts/arbitrage/BancorArbitrage.sol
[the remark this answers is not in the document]
barakman: Other than via designated payable functions?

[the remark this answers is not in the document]
yudilevi: yup - the arb contract trades, so it may receive ETH during a call to `trade` on any exchange

contracts/arbitrage/BancorArbitrage.sol
barakman: Maybe pass `customInt` as `uint24` then.

yudilevi: Can’t - the idea is that it’s a custom field that any exchange can use differently - one might use as uint24 and another might use it as unit256

test/arbitrage/BancorArbitrage.ts
```diff
@@ -0,0 +1,358 @@
...
+
+                        await transfer(deployer, bnt, exchanges.address, AMOUNT.mul(10));
+                        await transfer(deployer, arbToken1, exchanges.address, AMOUNT.mul(10));
+                        await transfer(deployer, arbToken2, exchanges.address, AMOUNT.mul(10));
+
+                        const routes = [
+                            {
+                                exchangeId: ExchangeId.BancorV2,
+                                targetToken: arbToken1.address,
+                                minTargetAmount: 1,
```
barakman, line 264: Minor: can use `getAmountsOut` here (and below) if you wanna make the test more restrictive.

test/arbitrage/BancorArbitrage.ts
```diff
@@ -0,0 +1,358 @@
...
+    });
+
+    describe('arbitrage', () => {
+        beforeEach(async () => {
+            await transfer(deployer, bnt, masterVault.address, AMOUNT.mul(10_000));
+            await bancorArbitrage.setRewards(ArbitrageRewardsDefaults);
+        });
+
+        // get all exchange ids (omit their names)
+        const exchangeIds = Object.values(ExchangeId).filter((key) => !isNaN(Number(key)));
```
barakman, line 227: Very minor: `Number` works for floats as well, probably better to use `parseInt`.

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+            sourceToken.safeApprove(address(router), sourceAmount);
+
+            // build the path
+            address[] memory path = new address[](2);
+
+            // perform the trade
+            if (sourceToken.isNative()) {
+                path[0] = address(_weth);
+                path[1] = address(targetToken);
+                router.swapExactETHForTokens{ value: sourceAmount }(minTargetAmount, path, address(this), deadline);
```
barakman, line 386: Suggestion (for every usage of `swapExact`):
```
uint256[] memory amounts = router.swapExact...
require(amounts[0] == sourceAmount && amounts[1] >= minTargetAmount);
```

contracts/helpers/MockExchanges.sol
```diff
@@ -0,0 +1,119 @@
...
+
+import { Token } from "../token/Token.sol";
+import { TokenLibrary } from "../token/TokenLibrary.sol";
+
+contract MockExchanges {
+    using SafeERC20 for IERC20;
+    using TokenLibrary for Token;
+
+    // the address that represents the native token reserve
+    address private constant NATIVE_TOKEN_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
```
barakman, line 17: Can use the one at `TokenLibrary`.

barakman (changes requested):

test/arbitrage/BancorArbitrage.ts
yudilevi: done

contracts/helpers/MockExchanges.sol
yudilevi: it's private there

barakman (dismissed):

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+    ) external onlyAdmin validFee(settings.percentagePPM) greaterThanZero(settings.maxAmount) {
+        uint32 prevPercentagePPM = _rewards.percentagePPM;
+        uint256 prevMaxAmount = _rewards.maxAmount;
+
+        // return if the settings are the same
+        if (prevPercentagePPM == settings.percentagePPM && prevMaxAmount == settings.maxAmount) {
+            return;
+        }
+
+        _rewards = settings;
```
yarivbancor, line 230: setRewards -> updateSettings
 _rewards = settings > _settings = newSettings
 
 and so on everywhere else

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+     * @dev checks whether the specified number of routes is supported
+     */
+    modifier validRouteLength(Route[] calldata routes) {
+        // validate inputs
+        _validRouteLength(routes);
+
+        _;
+    }
+
+    function _validRouteLength(Route[] calldata routes) internal pure {
```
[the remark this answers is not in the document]
yudilevi: added a comment

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+        uint16 exchangeId;
+        Token targetToken;
+        uint256 minTargetAmount;
+        uint256 deadline;
+        address customAddress;
+        uint256 customInt;
+    }
+
+    // rewards settings
+    struct Rewards {
```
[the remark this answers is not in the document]
yudilevi: updated all to rewards to keep it explicit

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -0,0 +1,479 @@
...
+            return;
+        }
+
+        revert InvalidExchangeId();
+    }
+
+    /**
+     * @dev allocates the rewards to the caller and burns the rest
+     */
+    function allocateRewards(Route[] calldata routes, uint256 sourceAmount, address caller) internal {
```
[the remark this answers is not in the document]
yudilevi: updated

contracts/arbitrage/BancorArbitrage.sol
yudilevi: see comment above re. explicitness

test/arbitrage/BancorArbitrage.ts
```diff
@@ -0,0 +1,358 @@
...
+        // increase BNT liquidity by the growth factor a few times
+        for (let i = 0; i < 5; i++) {
+            await depositToPool(deployer, token, 1, network);
+        }
+
+        return { poolToken, token };
+    };
+});
+
+// Initialization:
```
yarivbancor, line 1119: can be removed at this point?

ivanzhelyazkov: Removed

PR #472, opened 2023-03-03
Arb contract integration

ivanzhelyazkov:
* Allow arb contract to chain flash loan and trade functions on V3 network
* Exempt arb contract from flashloan fees
* Exempt arb contract from trade fees
* Add upgrade scripts for `BancorArbitrage`, `BancorNetwork` and `PoolCollection`


test/arbitrage/BancorArbitrage.ts
```diff
@@ -1030,6 +1108,48 @@ describe('BancorArbitrage', () => {
             );
         });
 
+        it('should be able to arbitrage using Bancor V3', async () => {
```
yudilevi, line 1111: instead of a dedicated test, we can re-enable the ones we initially disabled

test/helpers/Factory.ts
```diff
@@ -290,7 +290,8 @@ const createNetwork = async (
     networkSettings: NetworkSettings,
     masterVault: MasterVault,
     externalProtectionVault: ExternalProtectionVault,
-    bntPoolToken: PoolToken
+    bntPoolToken: PoolToken,
+    arbContractAddress: string
```
yudilevi, line 294: should follow the same format as the others, not just a string address but use the specific BancorArbitrage type etc.

yudilevi (changes requested):

contracts/helpers/MockExchangesNegativeOutput.sol
```diff
@@ -0,0 +1,151 @@
+// SPDX-License-Identifier: SEE LICENSE IN LICENSE
+pragma solidity 0.8.13;
```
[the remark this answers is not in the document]
ivanzhelyazkov: Added amount and a flag in the constructor to control how much is gained / lost from each swap in the exchange.

contracts/network/BancorNetwork.sol
```diff
@@ -618,6 +624,50 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
             );
     }
 
+    /**
+     * @inheritdoc IBancorNetwork
+     */
+    function tradeBySourceAmountArb(
```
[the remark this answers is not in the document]
ivanzhelyazkov: Created the function.

test/arbitrage/BancorArbitrage.ts
ivanzhelyazkov: Removed that test - the others are enabled.

test/helpers/Factory.ts
ivanzhelyazkov: Discussed to leave it due to complications requiring to change all BancorNetwork deployment tests.

test/arbitrage/BancorArbitrage.ts
```diff
@@ -986,9 +1067,6 @@ describe('BancorArbitrage', () => {
         const uniV3Fees = [100, 500, 3000];
         const tokenSymbols = [TokenSymbol.TKN1, TokenSymbol.TKN2, TokenSymbol.ETH];
 
-        // remove BancorV3 exchange until the reentrancy guard issue is resolved
-        exchangeIds.splice(exchangeIds.indexOf(ExchangeId.BancorV3), 1);
```
yudilevi, line 990: exists in other places in the test file as well

yudilevi (changes requested):

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -367,8 +367,10 @@ contract BancorArbitrage is ReentrancyGuardUpgradeable, Utils, Upgradeable {
             // allow the network to withdraw the source tokens
             sourceToken.safeApprove(address(_bancorNetworkV3), sourceAmount);
 
+            uint256 val = sourceToken.isNative() ? sourceAmount : 0;
+
             // perform the trade
-            _bancorNetworkV3.tradeBySourceAmountArb(
+            _bancorNetworkV3.tradeBySourceAmountArb{ value: val }(
```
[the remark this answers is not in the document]
ivanzhelyazkov: Replaced the custom interface with the actual `IBancorNetwork`.

contracts/helpers/MockExchanges.sol
```diff
@@ -136,16 +144,26 @@ contract MockExchanges {
...
-    ) public returns (uint256) {
+    ) public payable returns (uint256) {
         require(deadline >= block.timestamp, "Swap timeout");
         // withdraw source amount
         sourceToken.safeTransferFrom(trader, address(this), amount);
 
         // transfer target amount
-        // receive 1 token per swap
-        uint256 targetAmount = amount + 1e18;
+        // receive gainAmount tokens per swap
```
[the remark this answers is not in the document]
ivanzhelyazkov: Changed comment.

contracts/helpers/MockExchanges.sol
```diff
@@ -136,16 +144,26 @@ contract MockExchanges {
...
+        uint targetAmount;
+        if(_profit) {
+            targetAmount = amount + _outputAmount;
+        } else {
+            targetAmount = amount - _outputAmount;
+        }
         require(targetAmount >= minTargetAmount, "InsufficientTargetAmount");
-        targetToken.safeTransfer(trader, targetAmount);
+        if(address(targetToken) == NATIVE_TOKEN_ADDRESS) {
```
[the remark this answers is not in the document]
ivanzhelyazkov: Left just `targetToken.safeTransfer`.

contracts/helpers/MockExchanges.sol
```diff
@@ -136,16 +144,26 @@ contract MockExchanges {
         address trader,
         uint deadline,
         uint minTargetAmount
-    ) public returns (uint256) {
+    ) public payable returns (uint256) {
```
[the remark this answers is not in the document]
ivanzhelyazkov: Changed the visibility to private.

contracts/helpers/TestWETH.sol
```diff
@@ -23,4 +23,4 @@ contract TestWETH is ERC20 {
         payable(msg.sender).transfer(amount);
         emit Withdrawal(msg.sender, amount);
     }
-}
+}
```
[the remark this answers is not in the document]
ivanzhelyazkov: Will commit all formatted files at the end of the reviews.

contracts/network/BancorNetwork.sol
```diff
@@ -221,14 +224,16 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         INetworkSettings initNetworkSettings,
         IMasterVault initMasterVault,
         IExternalProtectionVault initExternalProtectionVault,
-        IPoolToken initBNTPoolToken
+        IPoolToken initBNTPoolToken,
+        address bancorArbitrage
```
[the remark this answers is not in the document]
ivanzhelyazkov: Discussed to leave it as address, due to no functions being called on it, only the address is used in the `BancorNetwork` contract.

contracts/network/BancorNetwork.sol
```diff
@@ -948,6 +972,50 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
...
+        uint256 deadline,
+        address beneficiary
+    ) private returns (uint256) {
+        _verifyTradeParams(sourceToken, targetToken, sourceAmount, minReturnAmount, deadline);
+
+        return
+            _trade(
+                TradeTokens({ sourceToken: sourceToken, targetToken: targetToken }),
+                TradeParams({ bySourceAmount: true, amount: sourceAmount, limit: minReturnAmount }),
+                TraderInfo({ trader: msg.sender, beneficiary: beneficiary }),
```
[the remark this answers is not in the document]
ivanzhelyazkov: Fixed by passing msg.sender as an argument.

contracts/network/BancorNetwork.sol
```diff
@@ -1079,20 +1147,27 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
             ? TradeTokens({ sourceToken: Token(address(_bnt)), targetToken: pool })
             : TradeTokens({ sourceToken: pool, targetToken: Token(address(_bnt)) });
 
+        bool ignoreFees = false;
```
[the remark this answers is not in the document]
ivanzhelyazkov: Added to TradeParams and adjusted all functions.

deploy/scripts/000060-bancor-network-upgrade.ts
```diff
@@ -0,0 +1,62 @@
...
+    upgradeProxy
+} from '../../utils/Deploy';
+import { Roles } from '../../utils/Roles';
+import { DeployFunction } from 'hardhat-deploy/types';
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+
+const func: DeployFunction = async ({ getNamedAccounts }: HardhatRuntimeEnvironment) => {
+    const { deployer } = await getNamedAccounts();
+
+    await grantRole({
```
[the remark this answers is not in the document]
ivanzhelyazkov: Combined both the `BancorNetwork` and `PoolCollection` upgrades in one script - first pausing the network, upgrading it, then upgrading the `PoolCollection`, and resuming at the end.

deploy/scripts/000060-bancor-network-upgrade.ts
```diff
@@ -0,0 +1,62 @@
...
+        from: deployer
+    });
+
+    await execute({
+        name: InstanceName.BancorNetwork,
+        methodName: 'resume',
+        from: deployer
+    });
+
+    return true;
```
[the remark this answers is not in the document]
ivanzhelyazkov: Turned out the role was already set, so removed the `setRole` in the beginning.

deployments/mainnet/types/index.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable camelcase */
 
-export type { PoolCollection as PoolCollectionType1V9 } from './contracts/pools/PoolCollectionType1V9';
+export type { PoolCollection as PoolCollectionType1V11 } from './contracts/pools/PoolCollectionType1V11';
```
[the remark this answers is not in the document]
ivanzhelyazkov: Corrected version.

test/arbitrage/BancorArbitrage.ts
ivanzhelyazkov: Fixed - arb tests now use all exchanges.

test/arbitrage/BancorArbitrage.ts
```diff
@@ -220,14 +220,10 @@ describe('BancorArbitrage', () => {
 
         describe('distribution and burn', () => {
             // get all exchange ids (omit their names)
-            const exchangeIds = Object.values(ExchangeId).filter((key) => !isNaN(parseInt(key as string)));
```
yudilevi, line 222: this prevents double keys to be in the list of ids, you sure it's not needed?

test/arbitrage/BancorArbitrage.ts
```diff
@@ -529,6 +522,9 @@ describe('BancorArbitrage', () => {
             const uniswapV3RouterSameOutput = sameOutputExchanges;
             const sushiswapV2RouterSameOutput = sameOutputExchanges;
 
+            // exclude v3 due to new arb contract deployment
+            exchangeIds.splice(exchangeIds.indexOf(ExchangeId.BancorV3), 1);
```
yudilevi, line 526: why is this needed?

ivanzhelyazkov: We're deploying a new arb contract here with new mock exchanges, to test out the 0 reward and burn amount case.
We deploy a new contract, b/c the exchanges addresses are immutable, but when we deploy a new contract, we can't trade on the same v3 - because the arb contract is immutable in the network. 
So if we want to trade using the v3 network here, we need to make an entirely new deployment (which uses a different address for the arb contract, meaning we can't reuse the createSystem function).

test/arbitrage/BancorArbitrage.ts
ivanzhelyazkov: Yes, exchangeIds for the distribution and burn tests are not needed.

yudilevi (approved):

contracts/arbitrage/BancorArbitrage.sol
```diff
@@ -486,4 +474,18 @@ contract BancorArbitrage is ReentrancyGuardUpgradeable, Utils, Upgradeable {
...
+     * @dev set exchange allowance to the max amount if it's less than the input amount
+     */
+    function _setExchangeAllowance(Token token, address exchange, uint inputAmount) private {
+        if(token.isNative()) {
+            return;
+        }
+        uint allowance = token.toIERC20().allowance(address(this), exchange);
+        if(allowance < inputAmount) {
+            // increase allowance to the max amount if allowance < inputAmount
+            token.safeIncreaseAllowance(exchange, type(uint256).max - allowance);
```
barakman, line 488: First off, you can probably use `token.safeApprove(exchange, type(uint256).max)` here.

This would save the gas used inside `safeIncreaseAllowance` in order to read the current allowance.

Second, why not use `safeIncreaseAllowance(exchange, inputAmount - allowance)`?

Or following the first comment - even `token.safeApprove(exchange, inputAmount)`?

The first reason is that the user will get more gas back after `transferFrom` is executed, because it will reset the allowance to zero (rather than just reducing it to some non-zero value).

But the more important reason is that an unlimited allowance is a potential future problem, for example, a hack in this contract would "grant" the unlimited allowance to the hacker.

contracts/network/BancorNetwork.sol
```diff
@@ -607,15 +606,35 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
...
+     */
+    function tradeByTargetAmountArb(
+        Token sourceToken,
+        Token targetToken,
+        uint256 targetAmount,
+        uint256 maxSourceAmount,
+        uint256 deadline,
+        address beneficiary
+    ) external payable whenNotPaused only(_bancorArbitrage) returns (uint256) {
+        return _tradeByTargetAmount(sourceToken, targetToken, targetAmount, maxSourceAmount, deadline, beneficiary, msg.sender);
```
barakman, line 637: For both `tradeBySourceAmountArb` and `tradeByTargetAmountArb`:

Can anyone here thinks of a scenario where `nonReentrant` is required because `only(_bancorArbitrage)` somehow doesn't prevent reentrancy?

contracts/pools/PoolCollection.sol
```diff
@@ -1619,7 +1627,7 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
                 revert InsufficientSourceAmount();
```
barakman, line 1627: Coincidentally noticed this check, and it feels that either the comment is wrong, or the actual condition is wrong.

test/arbitrage/BancorArbitrage.ts
```diff
@@ -965,12 +978,79 @@ describe('BancorArbitrage', () => {
...
+                        minTargetAmount: 1,
+                        deadline: DEADLINE,
+                        customAddress: bnt.address,
+                        customInt: 0
+                    }
+                ];
+
+                // check that the fee amount is 0
+                // use different fees
+                for (const tradeFee of [0.03, 0.05, 1, 2, 3]) {
```
barakman, line 1034: Watch out as `0.03` doesn't have an accurate representation AFAIK.
Might be better pass an ASCII (string) representation of it.

barakman (changes requested):

contracts/pools/PoolCollection.sol
ivanzhelyazkov: Hmm, this wasn't a part of the code changes in this PR

contracts/arbitrage/BancorArbitrage.sol
barakman: The security issue is with regards to what we don't know, not with regards to what we know.
Any hack (which we are not aware of), could potentially enjoy the unlimited allowance.

For example:
```
_setExchangeAllowance(Token(tokenIn), address(_uniswapV3Router), sourceAmount);
```
How do we ensure that an upgraded version of `_uniswapV3Router` doesn't transfer the entire allowance out of our contract?

Also, if we do decide to go ahead with this approach, then why not give unlimited approval to all known recipients, immediate upon the construction (deployment) of this contract?
Then we don't need to do any more calls to read the allowance, let alone to write it.

contracts/network/BancorNetwork.sol
ivanzhelyazkov: Actually, with the current design, if nonReentrant is added to this function, the arbs won't work - since the arb contract does a flashloan + trade, the V3 network gets locked after the flashloan - and adding `nonReentrant` here will revert the arb.
One important point to be considered is that multiple trades in one call will be possible from the arb contract with this change - something which wasn't possible before.

contracts/pools/PoolCollection.sol
barakman: True, but let's just verify that there's no issue here.

test/arbitrage/BancorArbitrage.ts
ivanzhelyazkov: Outputted the resulting trade fee for each item here and the resulting PPM conversion- it seems okay.

contracts/pools/PoolCollection.sol
ivanzhelyazkov: It seems correct - 
this is a trade by target amount, and when calculating the estimated source amount needed to make the trade, we check if it's 0 (so source balance or fullTargetAmount is 0 in the calculation) or more than the initially provided `maxSourceAmount` to `tradeByTargetAmount` function (so user hasn't set enough `maxSourceAmount` to get that target amount).

barakman (dismissed):

yarivbancor (approved):

yudilevi (approved):

PR #487, opened 2023-09-07
Disable trading internally as part of surplus withdrawal

barakman:
During surplus withdrawal, if trading is not yet disabled, then enable it instead of reverting.


contracts/pools/PoolCollection.sol
```diff
@@ -282,8 +283,8 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
         _poolTokenFactory = initPoolTokenFactory;
         _poolMigrator = initPoolMigrator;
 
-        _setDefaultTradingFeePPM(DEFAULT_TRADING_FEE_PPM);
-        _setNetworkFeePPM(DEFAULT_NETWORK_FEE_PPM);
+        _defaultTradingFeePPM = (DEFAULT_TRADING_FEE_PPM);
```
lbeder, line 286: We have usually been emitting events during construction as well (since this is how the backend was built). Are you sure it's no longer required to be this way?

lbeder: (Or is it mostly to reduce the size of the contract?)

barakman: It is ONLY for that purpose.
I've figured - since this is not an upgradable contract, we may as well "don't care" that no events are emitted in this case (though I do understand that it might actually be a counter-argument for what I am trying to reason).

test/network/BancorNetwork.ts
```diff
@@ -1489,21 +1489,6 @@ describe('BancorNetwork', () => {
...
-            const token = await createToken(tokenData);
-            await createPool(token, network, networkSettings, poolCollection);
-
-            await networkSettings.setFundingLimit(token.address, MAX_UINT256);
-            await depositToPool(emergencyStopper, token, INITIAL_LIQUIDITY, network);
-
-            // enable trading
-            await poolCollection.enableTrading(token.address, BNT_VIRTUAL_BALANCE, BASE_TOKEN_VIRTUAL_BALANCE);
-            // expect revert
-            await expect(network.withdrawPOL(token.address)).to.be.revertedWithError('TradingEnabled');
```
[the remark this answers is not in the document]
barakman: Added, Tx

contracts/pools/PoolCollection.sol
lbeder: That's a "risky" approach because this will create a problem when moving to a new chain, where the "initial" event has never been seen by the backend. I'll let @yudilevi to revert on this.

barakman: OK, keep in mind that the context here is V3, which is not very likely to be migrated to another chain.

contracts/network/BancorNetwork.sol
```diff
@@ -609,12 +609,9 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         if (!_networkSettings.isTokenWhitelisted(pool)) {
             revert NotWhitelisted();
         }
+
         // verify pool collection exists and retrieve it
         IPoolCollection poolCollection = _poolCollection(pool);
-        // verify trading is disabled for pool
-        if (poolCollection.tradingEnabled(pool)) {
-            revert TradingEnabled();
```
ivanzhelyazkov, line 616: TradingEnabled is unused, can remove the error definition

ivanzhelyazkov (commented):
Looks good
Need to update withdrawPOL function description in IBancorNetwork.sol

contracts/pools/PoolCollection.sol
lbeder: @yudilevi WDYT?

test/pools/PoolCollection.ts
```diff
@@ -375,20 +375,18 @@ describe('PoolCollection', () => {
...
 
             expect(await poolCollection.poolType()).to.equal(PoolType.Standard);
             expect(await poolCollection.defaultTradingFeePPM()).to.equal(DEFAULT_TRADING_FEE_PPM);
             expect(await poolCollection.networkFeePPM()).to.equal(DEFAULT_NETWORK_FEE_PPM);
             expect(await poolCollection.protectionEnabled()).to.equal(false);
 
             await expect(poolCollection.deployTransaction)
-                .to.emit(poolCollection, 'DefaultTradingFeePPMUpdated')
-                .withArgs(0, DEFAULT_TRADING_FEE_PPM);
+                .to.not.emit(poolCollection, 'DefaultTradingFeePPMUpdated');
```
lbeder, line 386: Unnecessary, IMHO

barakman: Of course, I kinda left it as a "stub", to remind us that it should have been there.
Do you recommend removing it?

contracts/network/BancorNetwork.sol
barakman: Fixed.

barakman:
> Looks good Need to update withdrawPOL function description in IBancorNetwork.sol

Done

contracts/pools/PoolCollection.sol
yudilevi: I'd try to avoid breaking the convention. Do keep in mind this contract is re-deployed every upgrade (including this one) so these settings will actually get executed and the events will be missing from the live backend

contracts/network/BancorNetwork.sol
```diff
@@ -609,12 +608,9 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         if (!_networkSettings.isTokenWhitelisted(pool)) {
```
yudilevi, line 608: the new whitelist var/check is completely missing

contracts/pools/PoolCollection.sol
```diff
@@ -595,6 +610,17 @@ contract PoolCollection is IPoolCollection, Owned, BlockNumber, Utils {
         _resetTradingLiquidity(bytes32(0), pool, data, data.liquidity, TRADING_STATUS_UPDATE_ADMIN);
     }
 
+    /**
+     * @inheritdoc IPoolCollection
+     */
+    function disableTradingByNetwork(Token pool) external only(address(_network)) {
```
yudilevi, line 616: this skips the entire logic associated with disabling trading on pools - reducing trading liquidity, emitting the associated events etc.
given these implications, I think the better approach would be to allow the network to execute the current enableTrading method

yudilevi (changes requested):

contracts/network/BancorNetwork.sol
barakman: Added

contracts/pools/PoolCollection.sol
barakman: Fixed

contracts/pools/PoolCollection.sol
barakman: The purpose is reducing the size of the contract

yudilevi: it will likely break the existing backend - any workaround?

barakman: Fixed

ivanzhelyazkov (dismissed):
LGTM (Only need to fix lint issues)

yudilevi (approved):

PR #489, opened 2023-10-19
burn network fees

ivanzhelyazkov:
* Replace network fee withdrawal with network fee burn (public function)
* Add min burn amount, initially set to 1M BNT
* Add tests

### DIFF
--- a/contracts/network/BancorNetwork.sol
+++ b/contracts/network/BancorNetwork.sol
@@ -95,9 +95,6 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
     // the emergency manager role is required to pause/unpause the network
     bytes32 private constant ROLE_EMERGENCY_STOPPER = keccak256("ROLE_EMERGENCY_STOPPER");
 
-    // the network fee manager role is required to pull the accumulated pending network fees
-    bytes32 private constant ROLE_NETWORK_FEE_MANAGER = keccak256("ROLE_NETWORK_FEE_MANAGER");
-
     // the address of the BNT token
     IERC20 private immutable _bnt;
 
@@ -149,15 +146,18 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
     // a mapping between pools and their respective pool collections
     mapping(Token => IPoolCollection) private _collectionByPool;
 
-    // the pending network fee amount to be burned by the vortex
+    // the pending network fee amount to be burned
     uint256 internal _pendingNetworkFeeAmount;
 
     bool private _depositingEnabled = true;
 
     uint32 private _polRewardsPPM;
 
+    // min network fee amount that can be burned
+    uint256 private _minNetworkFeeBurn;
+
     // upgrade forward-compatibility storage gap
-    uint256[MAX_GAP - 11] private __gap;
+    uint256[MAX_GAP - 12] private __gap;
 
     /**
      * @dev triggered when a new pool collection is added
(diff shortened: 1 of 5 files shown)


contracts/network/BancorNetwork.sol
```diff
@@ -149,15 +146,18 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
...
-    // the pending network fee amount to be burned by the vortex
+    // the pending network fee amount to be burned
     uint256 internal _pendingNetworkFeeAmount;
 
     bool private _depositingEnabled = true;
 
     uint32 private _polRewardsPPM;
 
+    // min network fee amount that can be burned
+    uint256 private _minNetworkFeeBurn;
```
lbeder, line 157: I wonder if we really need this. I.e., the caller can opt-in to check whether the accrued fee justifies the gas costs, even if it's only 1 wei. In any case, in the current implementation, the caller will still spend gas for naught, while the `burnNetworkFees` function gracefully returns

contracts/network/BancorNetwork.sol
```diff
@@ -820,26 +818,17 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
...
         uint256 currentPendingNetworkFeeAmount = _pendingNetworkFeeAmount;
-        if (currentPendingNetworkFeeAmount == 0) {
+        if (currentPendingNetworkFeeAmount < _minNetworkFeeBurn) {
             return 0;
         }
 
         _pendingNetworkFeeAmount = 0;
 
-        _masterVault.withdrawFunds(Token(address(_bnt)), payable(recipient), currentPendingNetworkFeeAmount);
+        _masterVault.withdrawFunds(Token(address(_bnt)), payable(address(_bnt)), currentPendingNetworkFeeAmount);
```
lbeder, line 829: I'd add a comment here that this "transfer to the token" burns the tokens (I do remember this, but it always seems weird on the first sight)

contracts/network/BancorNetwork.sol
```diff
@@ -315,11 +320,11 @@ contract BancorNetwork is IBancorNetwork, Upgradeable, ReentrancyGuardUpgradeabl
         // set up administrative roles
         _setRoleAdmin(ROLE_MIGRATION_MANAGER, ROLE_ADMIN);
         _setRoleAdmin(ROLE_EMERGENCY_STOPPER, ROLE_ADMIN);
-        _setRoleAdmin(ROLE_NETWORK_FEE_MANAGER, ROLE_ADMIN);
 
         _depositingEnabled = true;
 
         _setPOLRewardsPPM(2000);
+        _setMinNetworkFeeBurn(1_000_000e18);
```
lbeder, line 327: Since this also needs to be called during an upgrade, I'd consider adding the upgrade deployment script to this PR as well

contracts/network/BancorNetwork.sol
ivanzhelyazkov: This was part of the product spec. I'll let @yudilevi answer.

contracts/network/BancorNetwork.sol
ivanzhelyazkov: Added

contracts/network/BancorNetwork.sol
ivanzhelyazkov: Added the upgrade script

contracts/network/BancorNetwork.sol
ivanzhelyazkov: Discussed with Yudi, it was actually a DAO decision.

lbeder: Roger

lbeder (approved):

yudilevi (approved):

